PR #58, opened 2023-01-21
Simple controller: prove stability

marshtompsxd:
Enhance the liveness property of the simple controller by adding stability. That is, existence of the CR leads to existence of the configmap, and the configmap always exists from then.


marshtompsxd:
The client (https://github.com/vmware-research/verifiable-controllers/blob/main/src/kubernetes_cluster/spec/client.rs) only creates/deletes CR but does not touch intermediate objects.
The client might delete CR, but our liveness property starts with "if CR always exists..."

lalithsuresh:
LGTM

PR #60, opened 2023-02-07
Make ReconcileState a generic type RS

marshtompsxd:
ReconcileState was used to describe the local state of each Reconcile round and was previously a hardcoded type.
This PR makes it a generic type called T.

Signed-off-by: Xudong Sun <xudongs1@vmware.com>

### DIFF
--- a/src/kubernetes_cluster/proof/controller_runtime_liveness.rs
+++ b/src/kubernetes_cluster/proof/controller_runtime_liveness.rs
@@ -22,67 +22,67 @@ use builtin_macros::*;
 
 verus! {
 
-pub proof fn lemma_pre_leads_to_post_by_controller<RS>(reconciler: Reconciler<RS>, input: ControllerActionInput, action: ControllerAction<RS>, pre: StatePred<State<RS>>, post: StatePred<State<RS>>)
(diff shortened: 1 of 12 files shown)


lalithsuresh (commented):
Not sure I follow the types here. RS is ReconcileState, and then we have State<RS> (state of type ReconcileState), and then StatePred<State<RS>> (state predicate of type state of type ReconcileState). Better to replace RS with T to make it clear it's a generic state with opaque internals.

Now that said, given that the vast majority of the generic code does not need to care about the internals of ReconcileState, this would be something that is better handled with an associated type so that the generic parameters don't bubble throughout the code. Does Verus support associated types?

marshtompsxd:
```
error: The verifier does not yet support the following Rust feature: unsupported item
  --> kubernetes_cluster/spec/common.rs:12:1
   |
12 | / trait T {
13 | |     type A;
14 | |     type B;
15 | | }
   | |_^

error: aborting due to previous error
```
Seems not supported.
With associated types maybe I can make the current reconcile (a struct whose members are all closures) a trait and make ReconcileState (RS) a inner type of this trait.

PR #62, opened 2023-02-13
Implement simple controller based on the example from kube-rs

marshtompsxd:
This PR implements an untrusted simple controller that creates a configmap for the custom resource (which matches the spec). It is written in the "state machine" style -- the logic is written in a `reconcile_core` function which is called repeatedly in `reconcile` function. The `reconcile_core` takes the key to the custom resource object, the current local state, the response of the last request, and returns a new local state and a pending request.

Signed-off-by: Xudong Sun <xudongs1@vmware.com>


src/simple_controller_impl/src/main.rs
```diff
@@ -0,0 +1,204 @@
...
+    name: String,
+    namespace: String,
+    // kind: String,
+}
+
+// This is the API request that reconcile_core returns to reconcile
+// I did not find a way to encode the type into the enum
+// so each request operation is hardcoded to one type
+// This is a big limitation and we should work on it
+enum APIRequest {
```
lalithsuresh, line 50: Could you elaborate on this?

src/simple_controller_impl/src/main.rs
```diff
@@ -0,0 +1,204 @@
...
+#[tokio::main]
+async fn main() -> Result<()> {
+    tracing_subscriber::fmt::init();
+    let client = Client::try_default().await?;
+
+    let crs = Api::<SimpleCR>::all(client.clone());
+
+    info!("starting simple-controller");
+
+    Controller::new(crs, ListParams::default()) // The controller's reconcile is triggered when a CR is created/updated
```
lalithsuresh, line 192: The set up here with triggers determines when `reconcile_core()` is invoked. That is, this is a key assumption in our specs. Any thoughts on how to link it?

src/simple_controller_impl/src/main.rs.reference
```diff
@@ -0,0 +1,112 @@
+// Nightly clippy (0.1.64) considers Drop a side effect, see https://github.com/rust-lang/rust-clippy/issues/9608
```
lalithsuresh, line 1: I assume this is the normally written controller code we are trying to match?

src/simple_controller_impl/src/main.rs
marshtompsxd: I want the APIRequest enum to be not hardcoded to any resource type and to take a generic type T:
```rust
enum APIRequest<T> {
    GetReq(...),
    CreateReq(...),
    DeleteReq(...),
    ...
}
```
then I need to way to tell the reconcile function which type of resource to get/create/delete.

src/simple_controller_impl/src/main.rs.reference
marshtompsxd: Yes. This reference impl is slightly more complex (for example it checks the return error of create to decide whether to return error or ok)

src/simple_controller_impl/src/main.rs
marshtompsxd: It should be doable. The difficulty is to match the trigger condition written in a predicate to the imperative method sequences that set up the controller here.

src/simple_controller_impl/src/main.rs
marshtompsxd: Maybe there is a way around it. I will try again

lalithsuresh (approved):

PR #65, opened 2023-02-16
Add unique id to request/response pair

marshtompsxd:
To match the response (from the k8s API) and the request (sent to the k8s API), this PR adds a req_id to each request and the k8s API state machine will attach the same id to the response after handling each request. This req_id only applies to <request, response> pairs, and watch event messages do not have any such id.

The req_id has to be unique to differentiate responses for different requests. The controller, k8s API and client state machine maintains a req_id variable respectively which always increments after sending out one request. An alternative is to let the top-level k8s cluster state machine maintain the req_id and pass it to each individual state machine, but that requires the k8s cluster state machine to know how many requests each individual state machine sends out in each step so that it can increment the id properly.

The liveness proof was broken after introducing the req_id and now it is fixed. An important difference is that we now have the `[]exists |msg| P(msg)` pattern (e.g., see `lemma_always_reconcile_get_cr_done_implies_pending_get_cr_req`) in the proof and sometimes we need to unfold the temporal predicate to continue the proof (e.g., in `lemma_after_get_cr_pc_leads_to_cm_always_exists`). There should be a way to simplify this process with some lemma.

Note that we have not changed the simple controller spec to have branches for different response results. I will leave that for the next PR.

### DIFF
--- a/src/kubernetes_cluster/spec/kubernetes_api/state_machine.rs
+++ b/src/kubernetes_cluster/spec/kubernetes_api/state_machine.rs
@@ -16,7 +16,7 @@ verus! {
 
 // etcd is modeled as a centralized map that handles get/create/delete
 // TODO: support list/update/statusupdate
-pub open spec fn transition_by_etcd(msg: Message, s: KubernetesAPIState) -> (KubernetesAPIState, Message, KubernetesAPIActionInput)
+pub open spec fn transition_by_etcd(msg: Message, s: KubernetesAPIState) -> (EtcdState, Message, KubernetesAPIActionInput)
     recommends
         msg.content.is_APIRequest(),
 {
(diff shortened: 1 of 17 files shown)


src/kubernetes_cluster/spec/kubernetes_api/state_machine.rs
```diff
@@ -43,15 +44,15 @@ pub open spec fn transition_by_etcd(msg: Message, s: KubernetesAPIState) -> (Kub
...
+        let resp = form_msg(src, dst, list_resp_msg(result, req, req_id));
         (s_prime, resp, Option::None)
     } else if msg.is_create_request() {
         let req = msg.get_create_request();
         if s.resources.dom().contains(req.obj.key) {
             // Creation fails
             let s_prime = s;
             let result = Result::Err(APIError::ObjectAlreadyExists);
-            let resp = form_msg(src, dst, create_resp_msg(result, req));
+            let resp = form_msg(src, dst, create_resp_msg(result, req, req_id));
```
lalithsuresh, line 51: I'm not sure it's good to float req_id throughout the code like this. Is there no way for it to be assigned under the covers somehow in one spot?

marshtompsxd: I agree. The current way of assigning ids is a bit error-prone.
The best spot to assign the id is in the network state machine I think.
But we need to make sure the resp has the same id as the request, which can only be done at the k8s API state machine.

PR #68, opened 2023-03-01
Impl k8s openapi models

utaal:


src/resources/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+use crate::pervasive::string::*;
+
+// use k8s_openapi::api::core::v1::ConfigMap as K8SConfigMap;
+use k8s_openapi::apimachinery::pkg::apis::meta::v1::ObjectMeta as K8SObjectMeta;
+
+verus! {
+
+#[verifier(external_body)]
+pub struct ObjectMeta {
+    inner: K8SObjectMeta,
```
lalithsuresh, line 13: Are there restrictions on working with `inner` from the methods below (impl ObjectMeta....)?

utaal: Yes, only trusted `#[verifier(external_body)]` functions can work with `inner`, which is why verified code will have to access it through trusted functions (like `name`).

marshtompsxd:
I will work on top of this PR

marshtompsxd:
Move to #74 and close this one

PR #69, opened 2023-03-01
Use `==` for spec eq

marshtompsxd:
Since Verus supports ghost `==` https://github.com/verus-lang/verus/pull/352, we can stop using `===` and start to use `==` in spec and proof code. This PR simply replaces all the `===` with `==` and adds a few type annotations as Verus requires.


lalithsuresh:
LGTM

PR #71, opened 2023-03-05
Specify the controller's list-then-watch pattern

marshtompsxd:
Specify the controller's list-then-watch pattern.

Controllers use the list-then-watch pattern to observe the cluster state: it first issues a list to read the current cluster state, and the list returns a resource version indicating the "version" of the cluster state the controller just reads. Every single update to any object in the cluster state will increment the resource version number. After the list returns, the controller issues a watch request to continuously receive updates to the cluster state. When issuing the watch request, the controller specifies the resource version number returned by the list, which asks the apiserver to stream updates starting from the resource version.

The current specification is much more simplified than the above: since we have not specified resource version, list will not return any resource version. The watch request is also not specified yet so after the initial list returns the controller will enter a watching state and starts to receive whatever update notifications that are pending in the network.

This liveness proof of the simple controller is broken since the controller spec changes a lot. I will leave the proof there before the spec gets stable and fix all the broken proofs in one batch later.

### DIFF
--- a/src/kubernetes_cluster/spec/controller/controller_runtime.rs
+++ b/src/kubernetes_cluster/spec/controller/controller_runtime.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: MIT
 #![allow(unused_imports)]
 use crate::kubernetes_cluster::spec::{common::*, controller::common::*, reconciler::*};
-use crate::pervasive::{map::*, multiset::*, option::*, seq::*};
+use crate::pervasive::{map::*, multiset::*, option::*, seq::*, set::*};
 use crate::state_machine::action::*;
 use crate::state_machine::state_machine::*;
 use crate::temporal_logic::defs::*;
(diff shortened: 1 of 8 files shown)


src/kubernetes_cluster/spec/controller/controller_runtime.rs
```diff
@@ -11,36 +11,113 @@ use builtin_macros::*;
...
-/// It is highly simplified compared to the actual watcher implementation in kube-rs:
-/// (1) The triggering condition should come from the developer. Currently we hardcode it in
-/// reconcile_trigger.
-/// (2) The watcher and reconciler run concurrently: the watcher can take incoming events while
-/// the reconciler is in reconcile. If the reconciler is working on object X, all the incoming
-/// events related to X will stay pending in a queue.
-/// (3) The watcher deduplicates triggering events for the same object: if an event for object X
-/// comes but there is already another pending event for the same object, the incoming event is
-/// discarded.
+pub open spec fn issue_initial_list<T>(reconciler: Reconciler<T>) -> ControllerAction<T> {
```
lalithsuresh, line 14: Where is this used?

lalithsuresh (commented):
It would be good to start working on a shim API and tie the spec methods/proofs to that shim already, just so that we get a feeling for what each specification is contributing towards.

src/kubernetes_cluster/spec/controller/controller_runtime.rs
marshtompsxd: I forgot to add the new actions to the state machine.
They are used here: https://github.com/vmware-research/verifiable-controllers/pull/71/files#diff-5a79ed1c20cfc0ea5794df6915706fa337a25076d5c905da54d1fe9cc56f6235R40

marshtompsxd:
Yes. I will be working on getting some basic resource types' specs done and connecting them to the simple controller impl in a shim layer.

PR #74, opened 2023-03-07
Implement exec and spec version of k8s openapi objects

marshtompsxd:
Start to implement the exec and spec version of a few k8s openapi objects, including Pod, ConfigMap, PersistentVolumeClaim, StatefulSet. The exec implementation is a wrapper around the original implementation from k8s_openapi, and the spec version is a "view" of each corresponding exec implementation.

This early implementation is not complete and is just for building some toy controllers; there are many fields we need to specify later to support a real controller.


marshtompsxd:
@utaal What would be the best way to implement the finalizers() method for [finalizers](https://github.com/vmware-research/verifiable-controllers/pull/74/files#diff-79fa695244616ee6d5f45db7ab802223cc0c6f256717f44a2447883ca11b4b1eR23)? The spec version of finalizers is a `Seq<StringView>` and the exec version should be `Vec<String>` I think.

src/resources/service.rs
```diff
@@ -0,0 +1,132 @@
...
+use crate::resources::object_meta::*;
+
+use k8s_openapi::api::core::v1::Service as K8SService;
+use k8s_openapi::api::core::v1::ServiceSpec as K8SServiceSpec;
+use k8s_openapi::api::core::v1::ServiceStatus as K8SServiceStatus;
+
+verus! {
+
+#[verifier(external_body)]
+pub struct Service {
```
lalithsuresh, line 13: Do you already see yourself using the `Service` object type from a controller you want to build?

marshtompsxd: Yes. `Service` is often used to expose to the end user a stable IP of the back-end application managed by the controller.

utaal:
> @utaal What would be the best way to implement the finalizers() method for [finalizers](https://github.com/vmware-research/verifiable-controllers/pull/74/files#diff-79fa695244616ee6d5f45db7ab802223cc0c6f256717f44a2447883ca11b4b1eR23)? The spec version of finalizers is a `Seq<StringView>` and the exec version should be `Vec<String>` I think.

That sounds right. Let me know if you run into issues with that approach.

marshtompsxd:
> > @utaal What would be the best way to implement the finalizers() method for [finalizers](https://github.com/vmware-research/verifiable-controllers/pull/74/files#diff-79fa695244616ee6d5f45db7ab802223cc0c6f256717f44a2447883ca11b4b1eR23)? The spec version of finalizers is a `Seq<StringView>` and the exec version should be `Vec<String>` I think.
> 
> That sounds right. Let me know if you run into issues with that approach.

So I was trying to specify the view of the return value `ret` (whose type is `Vec<String>`) equals to the finalizers field `Seq<StringView>` in the ObjectMataView. But the problem is that `ret@` is `Seq<String>`, not `Seq<StringView>`.

marshtompsxd:
It seems to me we need an opaque function to return a Seq<StringView> for a Vec<String>, which says the Seq has the same size as the Vec, and seq[i] is the view of vec[i]

marshtompsxd:
We can work on finalizers later

PR #78, opened 2023-03-10
Use k8s openapi objects in spec and proof

marshtompsxd:
Replace all the toy k8s object definitions with the newly implemented k8s openapi objects in spec and proof.

### DIFF
--- a/src/kubernetes_api_objects/custom_resource.rs
+++ b/src/kubernetes_api_objects/custom_resource.rs
@@ -4,34 +4,29 @@ use crate::kubernetes_api_objects::common::*;
 use crate::kubernetes_api_objects::object_meta::*;
 use crate::pervasive::prelude::*;
 
-use k8s_openapi::api::apps::v1::StatefulSet as K8SStatefulSet;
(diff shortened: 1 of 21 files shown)


src/kubernetes_api_objects/custom_resource.rs
```diff
@@ -0,0 +1,140 @@
...
+use crate::kubernetes_api_objects::common::*;
+use crate::kubernetes_api_objects::object_meta::*;
+use crate::pervasive::prelude::*;
+
+verus! {
+
+// TODO: CustomResource should be a generic type
+#[verifier(external_body)]
+pub struct CustomResource {
+    // the content is specific to the controller
```
lalithsuresh, line 12: Is there no content (e.g. specific spec fields) in the CR that we need right now?

marshtompsxd: Not for this PR. After this PR we can start with putting fields such like `replica` into CR and prove the same number of pods eventually are set up.

lalithsuresh (approved):

PR #79, opened 2023-03-20
Assume the controller's reconcile is infinitely frequently triggered

marshtompsxd:
Add an action that checks if a cr exists in the Kubernetes API and, if so, inserts a reconcile request for that cr to the controller and assumes weak fairness on this action.

This allows us to set up the assumption that reconcile is infinitely frequently scheduled without modeling all the details about list-then-watch and resource versions.

Some of the existing lemmas for proving `p ~> reconcile_triggered` become deprecated and we will need to write new lemmas by packing the assumption into the spec and prove `true ~> reconcile_triggered`.


lalithsuresh:
Awesome patch. LGTM!

PR #82, opened 2023-03-23
Implement executable simple controller and prove refinement

marshtompsxd:
Implement the exec mode of the simple controller and prove it conforms to the specification.

The exec implementation requires wrapping around the exec mode of API requests, responses, and local states and their view for specifications. The refinement proof does not require any manual effort because we do not have any complex control flow (e.g., loop) yet.


lalithsuresh (commented):
What does the executable look like right now? Can it run with the shim layer?

marshtompsxd:
We don't have the shim layer yet; we will have to import kube-rs to write the shim layer. I will work on that when the new Verus release is out.

lalithsuresh:
LGTM

PR #86, opened 2023-03-29
Add channel manager to allocate channel id for each controller request

marshtompsxd:
Use ChannelManager to wrap around the channel id. ChannelManager is part of the host state machine and allocates a new channel id when sending a new request.

Ideally, the concept of channel should be modeled in the network state machine, but I found that will complicate the spec a lot. For example, we will need to differentiate whether the host is sending to an existing channel or sending to a new channel, and when a new channel is created in the network, the host somehow needs to know the channel id and waits for the message sent back to that channel. We will need to revisit the design of network state machine later.

### DIFF
--- a/src/kubernetes_cluster/spec/channel.rs
+++ b/src/kubernetes_cluster/spec/channel.rs
@@ -1,0 +1,28 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: MIT
+#![allow(unused_imports)]
+use crate::pervasive::prelude::*;
+use builtin::*;
+use builtin_macros::*;
+
+verus! {
+
+pub struct ChannelManager {
+    pub cur_chan_id: nat,
+}
+
+impl ChannelManager {
+    pub open spec fn init() -> Self {
+        ChannelManager {
+            cur_chan_id: 0,
+        }
+    }
+
+    pub open spec fn allocate(self) -> (Self, nat) {
+        (ChannelManager {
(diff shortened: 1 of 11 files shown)


src/kubernetes_cluster/spec/channel.rs
```diff
@@ -0,0 +1,34 @@
...
+
+// TODO: the concept of channel should be modeled inside the network state machine
+pub struct ChannelManager {
+    pub cur_chan_id: nat,
+}
+
+impl ChannelManager {
+    pub open spec fn init() -> Self {
+        ChannelManager {
+            cur_chan_id: 0,
```
lalithsuresh, line 21: Wouldn't we need to make sure there is only one global channel manager?

marshtompsxd: Right now we have one ChannelManager per each host state machine. Do you think it is better to have one single ChannelManager at the top-level state machine?

marshtompsxd: We need. We had some safety reasoning here: https://github.com/vmware-research/verifiable-controllers/blob/main/src/kubernetes_cluster/proof/controller_runtime_safety.rs#L32

The ChannelManager here just allocates a chan id whenever a request needs to be sent. It does not really split the network into channels. We still rely on the host state machine to correctly call `allocate`.

I thought about making the network state machine "channeled", but it seems hard and will complicate the spec a lot.

lalithsuresh:
LGTM

PR #88, opened 2023-03-31
Prove simple controller liveness except the reconcile error case

euclidgame:
Assume that `spec |= reconcile_ongoing ~> []cr_matched`, I complete the liveness proof. If we want to make `spec |= reconcile_ongoing ~> []cr_matched` verified as well, we have to prove `spec |= reconcile_error ~> []cr_matched`, since the other three cases (`init`, `get_cr`, `create_cm`) have all been proved. But considering it may not be possible to prove `spec |= reconcile_error ~> []cr_matched` under current specifications, another proof strategy is needed.

### DIFF
--- a/src/kubernetes_cluster/spec/distributed_system.rs
+++ b/src/kubernetes_cluster/spec/distributed_system.rs
@@ -21,7 +21,6 @@ use crate::pervasive::{map::*, option::*, seq::*, set::*};
 use crate::state_machine::action::*;
 use crate::state_machine::state_machine::*;
 use crate::temporal_logic::defs::*;
-use crate::temporal_logic::rules::*;
 use builtin::*;
 use builtin_macros::*;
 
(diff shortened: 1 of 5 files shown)


src/kubernetes_cluster/spec/distributed_system.rs
```diff
@@ -270,4 +271,27 @@ pub open spec fn controller_action_pre<T>(reconciler: Reconciler<T>, action: Con
     }
 }
 
+proof fn valid_stable_tla_forall_action_weak_fairness<T, Input, Output>(action: Action<State<T>, Input, Output>)
```
marshtompsxd, line 274: Can you create a `src/kubernetes_cluster/proof/cluster.rs` and move the two proofs to there?

src/simple_controller/proof/liveness.rs
```diff
@@ -50,13 +50,54 @@ proof fn liveness_proof_forall_cr()
...
+    });
+    valid_implies_implies_leads_to(sm_spec(simple_reconciler()), reconcile_ongoing_and_scheduled, lift_state(|s: State<SimpleReconcileState>| s.reconcile_state_contains(cr.object_ref())));
+    lemma_reconcile_ongoing_leads_to_cm_always_exists(cr);
+    leads_to_trans_temp(sm_spec(simple_reconciler()), reconcile_ongoing_and_scheduled, lift_state(|s: State<SimpleReconcileState>| s.reconcile_state_contains(cr.object_ref())), always(lift_state(cm_exists(cr))));
+    temp_pred_equality(reconcile_idle_and_scheduled.or(reconcile_ongoing_and_scheduled), lift_state(|s: State<SimpleReconcileState>| s.reconcile_scheduled_for(cr.object_ref())));
+    or_leads_to_combine_temp(sm_spec(simple_reconciler()), reconcile_idle_and_scheduled, reconcile_ongoing_and_scheduled, always(lift_state(cm_exists(cr))));
+}
 
 #[verifier(external_body)]
 proof fn lemma_reconcile_not_ongoing_leads_to_cm_always_exists(cr: CustomResourceView)
```
marshtompsxd, line 103: Feel free to remove it if not needed

src/simple_controller/proof/liveness.rs
```diff
@@ -74,16 +115,25 @@ proof fn lemma_reconcile_ongoing_leads_to_cm_always_exists(cr: CustomResourceVie
             lift_state(|s: State<SimpleReconcileState>| s.reconcile_state_contains(cr.object_ref()))
```
marshtompsxd, line 115: With the current spec we will not be able to prove this lemma. Instead, we will prove a different form of lemma under the assumption of []cr_exists. We can remove this lemma but keeping it for now is also fine.

src/kubernetes_cluster/spec/distributed_system.rs
euclidgame: Yes, I have done this in the latest commit.

PR #90, opened 2023-04-03
Move channel_manager to the top-level state machine

marshtompsxd:


src/kubernetes_cluster/proof/controller_runtime_liveness.rs
```diff
@@ -51,7 +51,7 @@ pub proof fn lemma_pre_leads_to_post_with_assumption_by_controller<T>(reconciler
     ensures
         sm_spec(reconciler).entails(lift_state(pre).and(always(lift_state(assumption))).leads_to(lift_state(post))),
 {
-    use_tla_forall::<State<T>, ControllerActionInput>(sm_spec(reconciler), |i| controller_next(reconciler).weak_fairness(i), input);
+    use_tla_forall::<State<T>, (Option<Message>, Option<ObjectRef>)>(sm_spec(reconciler), |i| controller_next(reconciler).weak_fairness(i), input);
```
lalithsuresh, line 54: Any particular reason you're unpacking the struct into a tuple with two elements?

marshtompsxd: We changed the input type of the top-level state machine's action (so it is no longer ControllerActionInput). I find the new input type is not used in many places so I just use a tuple instead of defining a new struct type.

PR #91, opened 2023-04-09
Specify controller crash behavior

marshtompsxd:
Specify controller crash and add weak fairness assumption to express "the controller eventually stops crashing".

To support specifying crash, we place controller_state in an Option and use Option::None to represent the case where the controller is not running, and we add a crash_enabled bit as the precondition to crash the controller. We add three new actions to the top-level state machine:
- crash_controller that sets the controller state to Option::None and if crash_enabled
- restart_controller that restarts the controller with the initial controller state if the controller is not running
- disable_crash that sets crash_enabled to false

We also add weak fairness assumption on restart_controller and disable_crash so that later we can prove eventually the controller stops crashing and is running alive.

Besides, we fix a few lemmas in cluster.rs. There are more broken lemmas to fix. Most of them need to be proved in a context where the controller is always alive and never crashes, which is not in the scope of this PR.

### DIFF
--- a/src/kubernetes_cluster/proof/cluster.rs
+++ b/src/kubernetes_cluster/proof/cluster.rs
@@ -28,6 +28,7 @@ use builtin_macros::*;
 
 verus! {
 
+/// Prove weak_fairness is stable.
 proof fn valid_stable_action_weak_fairness<T, Output>(action: Action<State<T>, (), Output>)
     ensures
         valid(stable(action.weak_fairness(()))),
@@ -36,6 +37,7 @@ proof fn valid_stable_action_weak_fairness<T, Output>(action: Action<State<T>, (
     always_p_stable::<State<T>>(split_always);
 }
 
+/// Prove weak_fairness for all input is stable.
 proof fn valid_stable_tla_forall_action_weak_fairness<T, Input, Output>(action: Action<State<T>, Input, Output>)
     ensures
         valid(stable(tla_forall(|input| action.weak_fairness(input)))),
@@ -45,6 +47,7 @@ proof fn valid_stable_tla_forall_action_weak_fairness<T, Input, Output>(action: 
     always_p_stable::<State<T>>(tla_forall(split_always));
 }
 
+/// Prove partial_spec is stable.
 pub proof fn valid_stable_sm_partial_spec<T>(reconciler: Reconciler<T>)
     ensures
         valid(stable(sm_partial_spec(reconciler))),
--- a/src/kubernetes_cluster/proof/cluster.rs
+++ b/src/kubernetes_cluster/proof/cluster.rs
@@ -28,6 +28,14 @@ use builtin_macros::*;
 
 verus! {
 
+proof fn valid_stable_action_weak_fairness<T, Output>(action: Action<State<T>, (), Output>)
+    ensures
+        valid(stable(action.weak_fairness(()))),
+{
+    let split_always = always(lift_state(action.pre(()))).implies(eventually(lift_action(action.forward(()))));
+    always_p_stable::<State<T>>(split_always);
+}
+
 proof fn valid_stable_tla_forall_action_weak_fairness<T, Input, Output>(action: Action<State<T>, Input, Output>)
     ensures
         valid(stable(tla_forall(|input| action.weak_fairness(input)))),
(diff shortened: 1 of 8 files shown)


euclidgame:
Do we need a weak fairness of disable_crash to ensure that the controller eventually stops crashing? I mean, is this requirement a little bit stronger or so?

marshtompsxd:
> Do we need a weak fairness of disable_crash to ensure that the controller eventually stops crashing? I mean, is this requirement a little bit stronger or so?

Without this weak fairness we won't be able to prove liveness. Think about an execution where the controller immediately crashes after each restart, and the controller will not be able to make any progress in that execution.

I think another potential way is to specify that eventually the interval between two crashes is large enough so that the controller can finish everything in between. But since we are just proving "eventually the desired state is achieved", that interval needs to be arbitrarily large.

euclidgame:
> Without this weak fairness we won't be able to prove liveness. 

Of course, I know this. What I wanna ask is can some spec like `<>disable_crash` replace the weak fairness?

marshtompsxd:
> can some spec like <>disable_crash replace the weak fairness?

euclidgame:
> so are you asking between `<>disable_crash` and `[]<>disable_crash` which one should we choose?

Yes, that’s what I mean.

marshtompsxd:
That's a good question. We will probably need sth like
```
pub open spec fn weaker_than_weak_fairness(self, input: Input) -> TempPred<State> {
    always(lift_state(self.pre(input))).implies(eventually(lift_action(self.forward(input))))
}
```
The good news is that no matter which way we go most of the proof is the same (the goal is to prove `[]next /\ []fair /\ []!crash_enabled /\ []cr_exists` |= true ~> []cr_match). The only difference is how to set up `spec |= []!crash_enabled`, which should be easy either way.

We should also discuss this with other members.

lalithsuresh:
This is probably something we want to discuss in depth.

> crash_controller that sets the controller state to Option::None and if crash_enabled

The Option::None feels like an odd way to model it. Do you even need to explicitly model the crashed state as opposed to "the controller will not do anything if crash_enabled is true"?

marshtompsxd:
> The Option::None feels like an odd way to model it. Do you even need to explicitly model the crashed state as opposed to "the controller will not do anything if crash_enabled is true"?

I didn't get the second part. Should we make the controller always stay in the initial state until crash_enabled is set to false? In that way, we are assuming the controller always crashes at the beginning of the reconciliation but never crashes in the middle.

marshtompsxd:
> Are we assuming controllers are "fail stop" ? That is, when they crash, it is basically a new instance that replaces it without any local persistent state from a previous incarnation?

Yes

> If so, I think it would be sufficient to simply prevent the controller from being able to interact with the network if crashed=true. This will also correctly cover the controller crashing in the middle of a reconcile.

By preventing, do you mean that the top-level controller_next action is disabled when crashed=true? But then when crashed=false, the action gets enabled and then the "new controller instance" will continue from the local state of the crashed controller.

lalithsuresh:
> By preventing, do you mean that the top-level controller_next action is disabled when crashed=true? But then when crashed=false, the action gets enabled and then the "new controller instance" will continue from the local state of the crashed controller.

Almost. The idea is to have a sequence of controllers C_i. When a controller C_i has crashed=true, it never gets crashed=false ever again, which means it is cut off completely from the network. A new controller (say C_{i+1}) then starts from the initial state (this is in fact what the behavior with pods being relaunched will look like in k8s).

marshtompsxd:
> The idea is to have a sequence of controllers C. When a controller C_i has crashed=true, it never gets crashed=false ever again, which means it is cut off completely from the network. A new controller (say C_{i+1}) then starts from the initial state (this is in fact what the behavior with pods being relaunched will look like in k8s).

Could you elaborate more on the difference between using Option and using Seq here? It seems to me the only difference is that for Option we reuse the same variable for each incarnation of the controller, but for Seq we move to the next index for the next incarnation of the controller.

lalithsuresh:
If it's easy to use Option and reset the controller correctly without having that Option spill throughout the controller spec, then it's fine.

That said, the Seq approach allows you to easily extend to having multiple controllers and reasoning about things like controller HA and leader election.

marshtompsxd:
I think using either Option or Seq would not affect the controller spec. The only modification I made to the controller spec is to add a function `init_controller_state` so that the top-level state machine can set the controller to the initial state when restarting it.

lalithsuresh:
LGTM

PR #92, opened 2023-04-13
Migrate to Verus main_new

tchajed:
The only changes to `*.rs` files are to switch imports from `crate::pervasive` to `vstd`.

There was an `impl` for `Option` in `api_method.rs` which only worked because Verus used to put the `pervasive` module inside the crate being built; now that Verus uses a separate vstd crate, this no longer works.


marshtompsxd:
LGTM

PR #93, opened 2023-04-13
Make crash and restart atomic

marshtompsxd:
Collapse the previous `crash_controller` and `restart_controller` into one action. This would simplify the proof because we no longer need to carry the Option. Note that the new specification on crash still allows the execution that the controller restart does not immediately happen after the crash: not scheduling any controller action before `restart_controller` is the same as the controller is crashed and then restarted by `restart_controller`.


lalithsuresh:
LGTM

PR #98, opened 2023-04-27
Build simple controller execuable with a shim layer

marshtompsxd:
The trusted shim layer connects the verified `reconcile_core` to the kube APIs. The shim layer implements `reconcile`, which repeatedly calls `reconcile_core` until it returns either error or done. Every time after calling `reconcile_core`, the shim layer parses the outgoing request, sends the request through kube APIs, and calls `reconcile_core` with the response.

Currently, we implement `SimpleCR` in deps_hack because `derive` is not well supported by Verus yet.


euclidgame:
LGTM

PR #109, opened 2023-05-11
Fix kube version to 0.78.0

marshtompsxd:


marshtompsxd:
close as it does not help fix the CI issues

PR #117, opened 2023-05-20
Implement the Zookeeper controller with refinement proof

marshtompsxd:
Port the unverified Zookeeper controller (from #113 ) to Anvil and finish refinement proof.

### Zookeeper controller

The Zookeeper controller can set up a Zookeeper cluster running in a Kubernetes cluster. To do so, it does the following:
- Create the headless service that assigns stable IP addresses and DNS names to each Zookeeper server (running in pods)
- Create the configmap that stores the configuration data and environment variables used by the Zookeeper server
- Create the statefulset that manages the containerized Zookeeper servers (i.e., pods) and the mounted volumes (i.e., pvcs and configmaps)

The custom resource, ZookeeperCluster, currently has only one field `replica` that specifies the number of desired Zookeeper servers to run.

### Newly supported Kubernetes resource types

The efforts to write the Zookeeper controller and its refinement using Anvil are moderate, and most changes in this PR are made to support reasoning about Kubernetes resource types used by the Zookeeper controller. More concretely, this PR implements the definition of these resource types by wrapping around the external definition from k8s_openapi library and their ghost types. The newly supported resource types include:
- Services: used for exposing network applications by assigning IP addresses and DNS names
- Pods: used for deploying containers
- PersistentVolumeClaims: used for providing storage volumes for containers
- StatefulSets: used for managing stateful applications

This PR also implements a trait Marshalable to help convert different **ghost** resource types to DynamicObjectView by marshaling their contents into JSON Value.

### What's missing

Compared to the Zookeeper controller from #113 , the ported Zookeeper controller has not set the owner reference of each created object yet. Besides, several fields (e.g., readiness_probe, command) in the Pod definition have no corresponding ghost types because we have no plan to reason about them in the near future.


euclidgame (approved):
You can merge this PR.

PR #125, opened 2023-05-30
Revise the k8s API spec to check namespace when creating objects

marshtompsxd:
To get closer to how Kubernetes handles creation requests in reality, this PR makes two major changes: (1) the creation request will carry a namespace field, and (2) the k8s API will check whether the namespace of the provided object matches the namespace sent on the creation request; if not, the k8s API will reject the request with the error BadRequest.

Currently the k8s API state machine handles creation requests as follows:
(1) if the namespace set in the provided object does not match the namespace sent on the creation request, then return BadRequest error, else go to (2)
(2) if the object already exists, then return AlreadyExists error, else go to (3)
(3) creates the object

The current specification for how to handle creation requests is still not perfect yet because there are other error cases (e.g., setting an invalid name). We will cover more cases in the spec when we need to reason about them in the proof.

### DIFF
--- a/src/controller_examples/simple_controller/spec/reconciler.rs
+++ b/src/controller_examples/simple_controller/spec/reconciler.rs
@@ -82,7 +82,6 @@ pub open spec fn make_config_map(cr: CustomResourceView) -> ConfigMapView
     let config_map = ConfigMapView::default()
         .set_metadata(ObjectMetaView::default()
             .set_name(cr.metadata.name.get_Some_0() + new_strlit("-cm")@)
-            .set_namespace(cr.metadata.namespace.get_Some_0())
         )
         .set_data(Map::empty().insert(new_strlit("content")@, cr.spec.content));
     config_map
@@ -91,6 +90,7 @@ pub open spec fn make_config_map(cr: CustomResourceView) -> ConfigMapView
 pub open spec fn create_cm_req(cr: CustomResourceView) -> APIRequest
 {
     APIRequest::CreateRequest(CreateRequest{
+        namespace: cr.metadata.namespace.get_Some_0(),
         obj: make_config_map(cr).to_dynamic_object(),
     })
 }
--- a/src/kubernetes_cluster/spec/kubernetes_api/state_machine.rs
+++ b/src/kubernetes_cluster/spec/kubernetes_api/state_machine.rs
@@ -1,7 +1,9 @@
 // Copyright 2022 VMware, Inc.
 // SPDX-License-Identifier: MIT
 #![allow(unused_imports)]
-use crate::kubernetes_api_objects::{api_method::*, common::*, dynamic::*, error::*};
+use crate::kubernetes_api_objects::{
+    api_method::*, common::*, dynamic::*, error::*, object_meta::*,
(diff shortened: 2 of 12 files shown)


src/controller_examples/simple_controller/spec/reconciler.rs
```diff
@@ -82,7 +82,6 @@ pub open spec fn make_config_map(cr: CustomResourceView) -> ConfigMapView
     let config_map = ConfigMapView::default()
         .set_metadata(ObjectMetaView::default()
             .set_name(cr.metadata.name.get_Some_0() + new_strlit("-cm")@)
-            .set_namespace(cr.metadata.namespace.get_Some_0())
```
euclidgame, line 85: The `namespace` is not set because you don't want to compare it with the request's namespace in `handle_create_req`?

marshtompsxd: Yes.
Also, it is unnecessary to set namespace here: if you don't set it, k8s API will set it; if you set it wrongly, the request will be rejected.

euclidgame: > if you don't set it, k8s API will set it

When will k8s API set it?

marshtompsxd: Do you mean in real execution? After the k8s API server receives the creation request, either the k8s API server or etcd will fill the namespace if that is empty. I think it is likely etcd (https://pkg.go.dev/go.etcd.io/etcd/clientv3/namespace) but I am not entirely sure. But I am sure that after the creat request gets handled, if you get the object you will see the namespace set in the object.

euclidgame: > But I am sure that after the create request gets handled, if you get the object you will see the namespace set in the object.

I only want to confirm this.

src/kubernetes_cluster/spec/kubernetes_api/state_machine.rs
```diff
@@ -54,19 +56,24 @@ pub open spec fn handle_create_request(msg: Message, s: KubernetesAPIState) -> (
         msg.content.is_create_request(),
 {
     let req = msg.content.get_create_request();
-    if s.resources.dom().contains(req.obj.object_ref()) {
-        // Creation fails
+    if req.obj.metadata.namespace.is_Some() && req.namespace != req.obj.metadata.namespace.get_Some_0() {
```
euclidgame, line 59: But if you don't set namespace for the inner object, when will the state machine go into this branch?

marshtompsxd: If some one sets a wrong namespace in the inner object, it will go to this branch.
In our example controllers we never do so, so we should be good in proof.

marshtompsxd: Note that such error case branches are also used to capture the mistakes if someone sets a wrong namespace in their objects -- if they do so, they will never prove liveness

euclidgame (approved):

PR #126, opened 2023-06-01
Replace the nat reconcile pc of the simple controller with enum

euclidgame:


src/simple_controller
marshtompsxd, line 1: Can you git rm this binary?

euclidgame: Done

PR #127, opened 2023-06-02
Implement ResourceWrapper for more k8s api resources

euclidgame:


src/controller_examples/simple_controller/common/mod.rs
marshtompsxd, line 1: Why does this PR include this file again? It is added by your previous PR, no?

src/controller_examples/simple_controller/exec/reconciler.rs
marshtompsxd, line 1: Same question here. Seems that the changes from the previous PR appear here again?

PR #128, opened 2023-06-02
Implement resource wrapper for more types

euclidgame:


euclidgame:
@marshtompsxd , I once tried to wrap `from_dynamic_object` in a trait, but I found it hard to write the postcondition for it.

PR #132, opened 2023-06-04
Make from_dynamic_object and unmarshal return Result

euclidgame:
In case that they may receive an object or a value that is an unexpected type or misses certain fields or has fields with unexpected types, we make the two methods `from_dynamic_object()` and `marshal()` return `Result<Self, Error>`.

In exec code, just check whether the `try_parse` succeed. In spec code, we have to reason about different cases. Because I divide the error into two types, in addition to the way of converting k8s objects to dynamic objects, the `Error` returned in all the `else` branches (in the spec code `from_dynamic_object()`) present an alternative sequence of the two types of errors, which is rather tedious.

### DIFF
--- a/src/controller_examples/simple_controller/spec/custom_resource.rs
+++ b/src/controller_examples/simple_controller/spec/custom_resource.rs
@@ -78,7 +78,7 @@ impl CustomResource {
             let res = CustomResource { inner: parse_result.unwrap() };
             Result::Ok(res)
         } else {
-            Result::Err(ParseDynamicObjectError::Error)
+            Result::Err(ParseDynamicObjectError::ExecError)
         }
     }
 }
--- a/src/controller_examples/simple_controller/spec/custom_resource.rs
+++ b/src/controller_examples/simple_controller/spec/custom_resource.rs
@@ -125,11 +125,11 @@ impl ResourceView for CustomResourceView {
     }
 
     open spec fn from_dynamic_object(obj: DynamicObjectView) -> Result<CustomResourceView, ParseDynamicObjectError> {
-        let data_object = obj.data.get_Some_0();
+        let data_object = obj.data.get_Object_0();
         let data_spec_unmarshal = CustomResourceSpecView::unmarshal(data_object[spec_field()]);
         if !obj.data.is_Object() {
             Result::Err(ParseDynamicObjectError::UnexpectedType)
-        } else if !obj.data.get_Some_0().dom().contains(spec_field()) {
+        } else if !data_object.dom().contains(spec_field()) {
             Result::Err(ParseDynamicObjectError::MissingField)
         } else if data_spec_unmarshal.is_Err() {
             Result::Err(ParseDynamicObjectError::UnmarshalError)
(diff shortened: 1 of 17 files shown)


src/controller_examples/simple_controller/spec/custom_resource.rs
```diff
@@ -133,15 +141,56 @@ impl ResourceView for CustomResourceView {
...
-    open spec fn from_dynamic_object(obj: DynamicObjectView) -> CustomResourceView {
-        CustomResourceView {
-            metadata: obj.metadata,
-            spec: CustomResourceSpecView{
-                content: obj.data.get_Object_0()[spec_field()].get_Object_0()[spec_content_field()].get_String_0(),
-            },
-            status: if obj.data.get_Object_0()[status_field()].is_Null() {Option::None} else {Option::Some(CustomResourceStatusView{
-                echoed_content: obj.data.get_Object_0()[status_field()].get_Object_0()[status_echoed_content_field()].get_String_0(),
-            })},
+    open spec fn from_dynamic_object(obj: DynamicObjectView) -> Result<CustomResourceView, ParseDynamicObjectError> {
```
marshtompsxd, line 144: You might want to avoid super long nested if-else because it hurts readability.
Could you refactor the code in the following format:
```
if ... {
  Error1
} else if ... {
  Error2
} else {
  Ok(...)
}
```
Note that you can "prefetch" the content in the enum like `let x = y.get_Some_0()` before checking the variant like `y.is_Some()` -- it still makes sense as long as you use `x` in a branch where `y.is_Some()` is true.

euclidgame: No problem

src/controller_examples/simple_controller/spec/custom_resource.rs
```diff
@@ -133,15 +141,56 @@ impl ResourceView for CustomResourceView {
...
+                let data_status = obj_data[status_field()];
+                if data_spec.is_Object() && (data_status.is_Null() || data_status.is_Object()) {
+                    let obj_data_spec = data_spec.get_Object_0();
+                    if obj_data_spec.dom().contains(spec_content_field()) {
+                        let data_spec_content = obj_data_spec[spec_content_field()];
+                        if data_spec_content.is_String() {
+                            if data_status.is_Null() {
+                                let res = CustomResourceView {
+                                    metadata: obj.metadata,
+                                    spec: CustomResourceSpecView { content: data_spec_content.get_String_0(), },
```
marshtompsxd, line 158: Instead of parsing `obj_data_spec` here, probably checking `obj_data_spec.unmarshal().is_Ok()` and then `spec: obj_data_spec.unmarshal().get_Ok()` is better? You can just have one error type called UnmarshalError for the not OK branch of unmarshal.

src/kubernetes_api_objects/error.rs
```diff
@@ -16,4 +16,10 @@ pub enum APIError {
     Other
 }
 
+pub enum ParseDynamicObjectError {
+    MissingField,
+    UnexpectedType,
+    Error,
```
marshtompsxd, line 22: Better to have a more concrete name than `Error`

euclidgame: It is a general one because we don’t know the concrete error info of exec code.

euclidgame:
@marshtompsxd I refine `CustomResourceView`. You can first check whether it is okay.

marshtompsxd:
> @marshtompsxd I refine `CustomResourceView`. You can first check whether it is okay.

LGTM

PR #135, opened 2023-06-04
Add trusted lemma `spec_integrity_is_preserved_by_marshal` to each Kubernetes object

marshtompsxd:
This PR adds a trusted lemma `spec_integrity_is_preserved_by_marshal` to each Kubernetes object, which says that the spec field of each object, after marshal and unmarshal, remains unchanged.

The main purpose is to help simplify reasoning about any execution where the controller creates an object and later gets the object and makes decisions based on the object's spec field. This lemma can help prove that the object's spec remains the same as when the controller creates the object if no updates happen to the object.

The existing lemma `to_dynamic_preserves_integrity` seems can do the same job but actually not: every time when creating or updating the object, the k8s API state machine will set certain fields in metadata such as resource_version, so the DynamicObject stored in etcd is not exactly the same one that was sent on the controller's create/update request.

This is inspired by the change made by #132 .

### DIFF
--- a/src/controller_examples/simple_controller/spec/custom_resource.rs
+++ b/src/controller_examples/simple_controller/spec/custom_resource.rs
@@ -97,7 +97,18 @@ impl ResourceWrapper<deps_hack::SimpleCR> for CustomResource {
     }
 }
 
-impl CustomResourceView {}
+impl CustomResourceView {
+    pub closed spec fn marshal_spec(s: CustomResourceSpecView) -> Value;
+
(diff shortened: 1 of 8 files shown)


euclidgame:
> This lemma can help prove that the object's spec remains the same as when the controller creates the object **if no updates happen to the object**.

Updating the objects will change the spec?

marshtompsxd:
> > This lemma can help prove that the object's spec remains the same as when the controller creates the object **if no updates happen to the object**.
> 
> Updating the objects will change the spec?

Yes

euclidgame (approved):

PR #136, opened 2023-06-05
Update statefulset for zookeeper controller

euclidgame:
In `reconcile_core` of the zookeeper controller, after creating config map, first send a get request to check whether the statefulset already exists in etcd.

### DIFF
--- a/src/controller_examples/zookeeper_controller/exec/reconciler.rs
+++ b/src/controller_examples/zookeeper_controller/exec/reconciler.rs
@@ -166,27 +166,71 @@ pub fn reconcile_core(
             return (state_prime, req_o);
         },
         ZookeeperReconcileStep::AfterCreateConfigMap => {
-            let stateful_set = make_stateful_set(zk);
-            let req_o = Option::Some(KubeAPIRequest::CreateRequest(
-                KubeCreateRequest {
+            let req_o = Option::Some(KubeAPIRequest::GetRequest(
+                KubeGetRequest {
                     api_resource: StatefulSet::api_resource(),
+                    name: zk.name().unwrap(),
                     namespace: zk.namespace().unwrap(),
-                    obj: stateful_set.to_dynamic_object(),
                 }
             ));
             let state_prime = ZookeeperReconcileState {
-                reconcile_step: ZookeeperReconcileStep::Done,
+                reconcile_step: ZookeeperReconcileStep::AfterGetStatefulSet,
                 ..state
             };
             return (state_prime, req_o);
         },
+        ZookeeperReconcileStep::AfterGetStatefulSet => {
+            if resp_o.is_some() && resp_o.as_ref().is_get_response() {
+                let stateful_set = make_stateful_set(zk);
+                let get_sts_resp = resp_o.unwrap().into_get_response();
+                if get_sts_resp.is_ok() {
+                    // update
+                    let old_sts_o = get_sts_resp.unwrap();
+                    let update_sts = StatefulSet::from_dynamic_object(old_sts_o).set_spec(stateful_set.spec());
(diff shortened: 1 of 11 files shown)


src/kubernetes_api_objects/persistent_volume_claim.rs
```diff
@@ -78,7 +78,7 @@ impl PersistentVolumeClaim {
     #[verifier(external_body)]
     pub fn api_resource() -> (res: ApiResource)
         ensures
-            res@.kind == Kind::CustomResourceKind,
+            res@.kind == Kind::PersistentVolumeClaimKind,
```
marshtompsxd, line 81: Thanks for catching the copy-paste bug. I think we should just do `res@.kind == PersistentVolumeClaimView::kind()` (`kind()` should not take self, my bad).

Let's do it in the next PR.

src/shim_layer/mod.rs
```diff
@@ -178,6 +178,31 @@ where
                         },
                     }
                 },
+                KubeAPIRequest::UpdateRequest(update_req) => {
+                    let api = Api::<deps_hack::kube::api::DynamicObject>::namespaced_with(
+                        client.clone(), update_req.namespace.as_rust_string_ref(), update_req.api_resource.as_kube_ref()
+                    );
+                    let pp = PostParams::default();
+                    let obj_to_update = update_req.obj.into_kube();
+                    match api.replace(update_req.name.as_rust_string_ref(), &&pp, &obj_to_update).await {
```
marshtompsxd, line 187: Why `&&pp` here (I mean double `&`)?

src/controller_examples/zookeeper_controller/exec/reconciler.rs
```diff
@@ -166,27 +166,75 @@ pub fn reconcile_core(
...
             return (state_prime, req_o);
         },
+        ZookeeperReconcileStep::AfterGetStatefulSet => {
+            if resp_o.is_some() && resp_o.as_ref().unwrap().is_get_response() {
+                let stateful_set = make_stateful_set(zk);
+                let get_sts_resp = resp_o.unwrap().into_get_response().res;
+                if get_sts_resp.is_ok() {
+                    // update
+                    let old_sts_o = get_sts_resp.unwrap();
+                    let update_sts = StatefulSet::from_dynamic_object(old_sts_o);
```
marshtompsxd, line 189: Can you rename this `update_sts` to `found_sts`? And there is no need to define `old_sts_o` since you only use it once here.

src/controller_examples/zookeeper_controller/exec/reconciler.rs
```diff
@@ -166,27 +166,75 @@ pub fn reconcile_core(
...
+                    if update_sts.is_ok() {
+                        let mut update_sts = update_sts.unwrap();
```
marshtompsxd, line 191: Can you rename this `update_sts` to `new_sts`?

src/controller_examples/zookeeper_controller/spec/reconciler.rs
```diff
@@ -103,16 +103,80 @@ pub open spec fn reconcile_core(
...
             (state_prime, req_o)
         },
+        ZookeeperReconcileStep::AfterGetStatefulSet => {
+            let stateful_set = make_stateful_set(zk);
+            if resp_o.is_Some() && resp_o.get_Some_0().is_GetResponse() {
+                let get_sts_resp = resp_o.get_Some_0().get_GetResponse_0().res;
+                if get_sts_resp.is_Ok() {
+                    // update
+                    let old_sts_o = get_sts_resp.get_Ok_0();
+                    let update_sts_obj = StatefulSetView::from_dynamic_object(old_sts_o);
```
marshtompsxd, line 125: Similar comment as in the exec code

PR #140, opened 2023-06-06
Make kind() an associated function, instead of a method

marshtompsxd:
Previously kind() was implemented as a method, which is unnecessary because it does not need any object information. This PR changes it to an associated function of the ResourceView trait.


ZichengMa (approved):

src/controller_examples/zookeeper_controller/spec/reconciler.rs
```diff
@@ -105,8 +105,9 @@ pub open spec fn reconcile_core(
             let stateful_set = make_stateful_set(zk);
```
euclidgame, line 105: This line can be deleted

euclidgame (approved):

PR #144, opened 2023-06-14
Support update operations for rabbitmq controller

ZichengMa:
- In the official rabbitmq controller, only scale-up is supported. When user wants to scale down, rabbitmq controller will reject the request and return `nil` in reconciliation loop. Currently, in our controller, an error will be returned.

- For scale up, the official rabbitmq controller will execute a command `rabbitmq-queues rebalance all`. This command is trying to split different queues' leader to different nodes, which will utilize resources more efficiently. However, this is not necessary if we only want the cluster to perform right functionality.

### DIFF
--- a/src/controller_examples/rabbitmq_controller/exec/reconciler.rs
+++ b/src/controller_examples/rabbitmq_controller/exec/reconciler.rs
@@ -280,8 +280,7 @@ pub fn reconcile_core(rabbitmq: &RabbitmqCluster, resp_o: Option<KubeAPIResponse
                                 return (state_prime, req_o);
                         }
                     }
-                }
-                else if get_sts_resp.unwrap_err().is_object_not_found() {
+                } else if get_sts_resp.unwrap_err().is_object_not_found() {
                     // create
                     let req_o = Option::Some(KubeAPIRequest::CreateRequest(
                         KubeCreateRequest {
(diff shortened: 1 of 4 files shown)


src/controller_examples/rabbitmq_controller/exec/reconciler.rs
```diff
@@ -238,20 +238,90 @@ pub fn reconcile_core(rabbitmq: &RabbitmqCluster, resp_o: Option<KubeAPIResponse
...
+                                ));
+                                let state_prime = RabbitmqReconcileState {
+                                    reconcile_step: RabbitmqReconcileStep::AfterUpdateStatefulSet,
+                                    ..state
+                                };
+                                return (state_prime, req_o);
+                        }
+                    }
+                }
+                else if get_sts_resp.unwrap_err().is_object_not_found() {
```
marshtompsxd, line 284: Can you move `else if` to the above line? Make it `} else if xxxx`

src/controller_examples/rabbitmq_controller/spec/reconciler.rs
```diff
@@ -184,11 +184,96 @@ pub open spec fn reconcile_core(
...
+            };
+            (state_prime, req_o)
+        },
+        RabbitmqReconcileStep::AfterGetStatefulSet => {
+            if resp_o.is_Some() && resp_o.get_Some_0().is_GetResponse() {
+                let stateful_set = make_stateful_set(rabbitmq);
+                let get_sts_resp = resp_o.get_Some_0().get_GetResponse_0().res;
+                if get_sts_resp.is_Ok() {
+                    // update
+                    let found_stateful_set = StatefulSetView::from_dynamic_object(get_sts_resp.get_Ok_0());
```
marshtompsxd, line 206: Can you refactor it to make it look like this https://github.com/vmware-research/verifiable-controllers/pull/146/files#diff-1c325cd2e5693d6b1afd85bfce7464e62f93486d4dc5367b634b9a29a8194a22R145 ?

PR #149, opened 2023-06-26
Cosmetic changes in ZookeeperCluster

marshtompsxd:
Split the ZookeeperCluster and its ghost version into different folders and rename `replica` to `replicas`.

### DIFF
--- a/src/controller_examples/zookeeper_controller/exec/reconciler.rs
+++ b/src/controller_examples/zookeeper_controller/exec/reconciler.rs
@@ -10,6 +10,7 @@ use crate::pervasive_ext::string_map::StringMap;
 use crate::pervasive_ext::string_view::*;
 use crate::reconciler::exec::*;
 use crate::zookeeper_controller::common::*;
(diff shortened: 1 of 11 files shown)


src/controller_examples/zookeeper_controller/exec/reconciler.rs
```diff
@@ -10,6 +10,7 @@ use crate::pervasive_ext::string_map::StringMap;
 use crate::pervasive_ext::string_view::*;
 use crate::reconciler::exec::*;
 use crate::zookeeper_controller::common::*;
+use crate::zookeeper_controller::exec::zookeepercluster::*;
 use crate::zookeeper_controller::spec::reconciler as zk_spec;
 use crate::zookeeper_controller::spec::zookeepercluster::*;
```
ZichengMa, line 15: I think this line can be removed since we import spec::zkcluster in exec::zkcluster.

marshtompsxd: no we still need this

PR #151, opened 2023-06-27
Prove lemma_true_leads_to_always_the_object_in_reconcile_has_spec_as

marshtompsxd:
This lemma states that from any state, it leads to a state where from now it is always true that if the controller is reconciling for an object whose key is `zk.object_ref()`, that object's spec is the same as `zk.spec`. This lemma requires that in the spec we have `[] desired_state_is(zk)`.

The proof is not very straightforward, we need to apply the unpack rule and another lemma `lemma_true_leads_to_always_the_object_in_schedule_has_spec_as`.

### DIFF
--- a/src/controller_examples/zookeeper_controller/proof/liveness.rs
+++ b/src/controller_examples/zookeeper_controller/proof/liveness.rs
@@ -124,17 +124,196 @@ proof fn lemma_true_leads_to_always_the_object_in_schedule_has_spec_as(spec: Tem
 }
 
 // TODO: generalize this
-#[verifier(external_body)]
 proof fn lemma_true_leads_to_always_the_object_in_reconcile_has_spec_as(spec: TempPred<ClusterState>, zk: ZookeeperClusterView)
     requires
         zk.well_formed(),
         spec.entails(always(lift_action(next::<ZookeeperClusterView, ZookeeperReconcileState, ZookeeperReconciler>()))),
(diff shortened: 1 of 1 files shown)


src/controller_examples/zookeeper_controller/proof/liveness.rs
```diff
@@ -124,17 +124,196 @@ proof fn lemma_true_leads_to_always_the_object_in_schedule_has_spec_as(spec: Tem
...
+                true_pred(),
+                lift_state(|s: ClusterState| !s.reconcile_state_contains(zk.object_ref())),
+                lift_state(safety::the_object_in_reconcile_has_spec_as(zk))
+            );
+            leads_to_stable_temp(stable_spec_with_assumption, lift_action(stronger_next), true_pred(), lift_state(safety::the_object_in_reconcile_has_spec_as(zk)));
+        }
+    );
+
+    // By unpacking the conditions we will have: stable_spec |= []the_object_in_schedule_has_spec_as(zk) ~> []the_object_in_reconcile_has_spec_as(zk)
+    assume(valid(stable(stable_spec)));
```
euclidgame, line 287: `assume`?

marshtompsxd: Yes, I leave this proof as assumed for now. I find it quite cumbersome to prove stability so I plan to prove all the stability in the liveness proof later. You will find other such `assume()` in the liveness proof.

euclidgame (approved):

PR #154, opened 2023-06-28
Add e2e tests for zookeeper and rabbitmq controller

ZichengMa:
The e2e tests check whether the zookeeper (resp. rabbitmq) controller successfully brings up a zookeeper (resp. rabbitmq) cluster. It will be integrated into our CI pipeline.


e2e_test/src/rabbitmq_e2e.rs
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+
+    // create a rabbitmq cluster
+    let pp = PostParams::default();
+    let rabbitmq_api: Api<RabbitmqCluster> = Api::namespaced(client.clone(), "default");
+    let rabbitmq_name = "rabbitmq-test";
+    let rabbitmq_sts_name = format!("{}-server", &rabbitmq_name);
+    let rabbitmq_cm_name = format!("{}-server-conf", &rabbitmq_name);
+    let rabbitmq = RabbitmqCluster::new(
```
marshtompsxd, line 66: @ZichengMa I would say we should just submit the rabbitmqcluster cr yaml instead of creating the object here. For two reasons: (1) We kind of repeat the definition of the rabbitmqcluster that we already have in deps_hack. So we should just make the e2e test simple and agnostic of the cr definition. (2) the other testing tools that can generate different inputs usually generate a cr yaml, so it is better to align with their output for later use.

PR #158, opened 2023-06-30
Prove termination of zookeeper controller

euclidgame:
Prove termination of zookeeper controller that when controller will no longer be restarted (always crash is disabled), the controller will always eventually enter reconcile idle state. That is basically `spec |= true ~> reconcile_idle`.

The strategy is to prove for reach case of the controller steps, `spec |= step ~> reconcile_idle` holds and then combine them together. It's obvious that there is no pending request at initial step and for other step (except done and error) there is either pending request or corresponding response in flight. Then use `wf1` of kubernetes api to show that in flight request will also lead to a corresponding response and use `wf1` of controller to show that with a response the controller with enter the next state.


euclidgame:
I decide to simplify the proof in the next PR.

src/controller_examples/zookeeper_controller/proof/safety.rs
```diff
@@ -392,6 +395,1027 @@ pub open spec fn sts_update_request_msg_since(key: ObjectRef, rest_id: RestId) -
...
+                        if s.message_in_flight(s.pending_req_of(key)) {
+                            assert(s_prime.message_in_flight(s_prime.pending_req_of(key)));
+                        } else {
+                            assert(s_prime.message_in_flight(resp));
+                        }
+                    } else {
+                        assert(s_prime.message_in_flight(s_prime.pending_req_of(key)));
+                    }
+                }
+                Step::ClientStep(input) => {
```
marshtompsxd, line 760: Would the proof still go through without reasoning about ClientStep (similarly for the lemmas below)?

src/controller_examples/zookeeper_controller/proof/terminate.rs
```diff
@@ -50,10 +54,1781 @@ pub proof fn reconcile_eventually_terminates(spec: TempPred<ClusterState>, zk: Z
...
+        at_after_update_stateful_set_step(zk.object_ref())(s)
+        && s.reconcile_state_of(zk.object_ref()).pending_req_msg.is_Some()
+        && is_controller_update_request(s.pending_req_of(zk.object_ref()))
+        && (s.message_in_flight(s.pending_req_of(zk.object_ref()))
+        || exists |resp_msg: Message| {
+            #[trigger] s.message_in_flight(resp_msg)
+            && resp_msg_matches_req_msg(resp_msg, s.pending_req_of(zk.object_ref()))
+        })
+    };
+    temp_pred_equality::<ClusterState>(lift_state(pending_req_in_flight_or_resp_in_flight_at_after_update_stateful_set_step(zk.object_ref())), lift_state(at_after_update_stateful_set_step(zk.object_ref())).implies(lift_state(at_after_update_stateful_set_step_and_pending_req_in_flight_or_resp_in_flight)));
```
marshtompsxd, line 147: This is nice

src/controller_examples/zookeeper_controller/proof/safety.rs
euclidgame: No, I tried without it at first.

marshtompsxd:
LGTM

PR #159, opened 2023-06-30
Simplify reasoning about schedule_controller_reconcile

marshtompsxd:
Previously, proving wf1 for schedule_controller_reconcile is cumbersome because we need `desired_state_is` (or `desired_state_exists`) to satisfy this action's precondition, but `desired_state_is` is in the spec and cannot be directly used to satisfy the precondition when applying wf1.

This PR introduces a new lemma to "borrow" the conditions from the spec to apply wf1, which makes it simpler to reason about schedule_controller_reconcile.

### DIFF
--- a/src/kubernetes_cluster/proof/controller_runtime_eventual_safety.rs
+++ b/src/kubernetes_cluster/proof/controller_runtime_eventual_safety.rs
@@ -65,15 +65,9 @@ pub proof fn lemma_true_leads_to_always_the_object_in_schedule_has_spec_as<K: Re
 
     K::object_ref_is_well_formed();
 
-    lemma_always_desired_state_exists(spec, cr);
-    temp_pred_equality::<State<K, T>>(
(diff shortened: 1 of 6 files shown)


src/kubernetes_cluster/proof/controller_runtime_eventual_safety.rs
```diff
@@ -65,15 +65,9 @@ pub proof fn lemma_true_leads_to_always_the_object_in_schedule_has_spec_as<K: Re
 
     K::object_ref_is_well_formed();
 
-    lemma_always_desired_state_exists(spec, cr);
-    temp_pred_equality::<State<K, T>>(
-        lift_state(desired_state_exists(cr)), lift_state(schedule_controller_reconcile().pre(input))
+    controller_runtime_liveness::lemma_pre_leads_to_post_by_schedule_controller_reconcile_borrow_from_spec(
```
euclidgame, line 68: I think the `desired_state_is(cr)(s)` in `stronger_next` is no longer needed?

PR #161, opened 2023-07-02
Simplify the termination and safety proof of zookeeper

euclidgame:
The general idea to simplify the proof is to find common patterns of what happens in `reconcile_core`.

The current `reconcile_core` of zookeeper controller is simple, so that I can come up with a single pattern that the controller starts from a state, reads the response, then generates a request and enters the next state.

### DIFF
--- a/src/controller_examples/zookeeper_controller/proof/safety.rs
+++ b/src/controller_examples/zookeeper_controller/proof/safety.rs
@@ -941,14 +941,14 @@ pub proof fn lemma_always_pending_req_in_flight_or_resp_in_flight_at_after_get_s
     );
 }
 
-pub open spec fn pending_req_in_flight_or_resp_in_flight_at_after_create_headless_service_step(
-    key: ObjectRef
+pub open spec fn pending_req_in_flight_or_resp_in_flight_at_after_create_obj_step(
+    key: ObjectRef, step: ZookeeperReconcileStep
 ) -> StatePred<ClusterState>
     recommends
         key.kind.is_CustomResourceKind(),
 {
     |s: ClusterState| {
-        at_zookeeper_step(key, ZookeeperReconcileStep::AfterCreateHeadlessService)(s)
+        at_zookeeper_step(key, step)(s)
             ==> {
                 s.reconcile_state_of(key).pending_req_msg.is_Some()
                 && is_controller_create_request(s.pending_req_of(key))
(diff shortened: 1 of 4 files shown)


src/controller_examples/zookeeper_controller/proof/safety.rs
```diff
@@ -846,112 +738,17 @@ pub proof fn lemma_always_reconcile_init_implies_no_pending_req(
...
-        spec,
-        init::<ZookeeperClusterView, ZookeeperReconcileState, ZookeeperReconciler>(),
-        stronger_next,
-        invariant
-    );
-}
-
-pub open spec fn pending_req_in_flight_or_resp_in_flight_at_after_create_headless_service_step(
-    key: ObjectRef
+pub open spec fn pending_req_in_flight_or_resp_in_flight_at_step(
```
marshtompsxd, line 741: This is nice

src/controller_examples/zookeeper_controller/proof/safety.rs
```diff
@@ -961,415 +758,39 @@ pub open spec fn pending_req_in_flight_or_resp_in_flight_at_after_create_headles
...
-                    #[trigger] s.message_in_flight(resp_msg)
-                    && resp_msg_matches_req_msg(resp_msg, s.pending_req_of(key))
-                })
-            }
-    }
-}
-
-pub proof fn lemma_always_pending_req_in_flight_or_resp_in_flight_at_after_create_stateful_set_step(
-    spec: TempPred<ClusterState>, key: ObjectRef
+pub proof fn lemma_always_pending_req_in_flight_or_resp_in_flight_at_step(
```
marshtompsxd, line 761: It seems to me we can even make it generally applicable to other controllers. Not super sure but we can try that later.

euclidgame: I agree. But if you want to lift such lemma to controller runtime, then you have to add some postconditions to `reconcile_core`.

euclidgame:
The general idea to simplify the proof is to find common patterns of what happens in `reconcile_core`.

The current `reconcile_core` of zookeeper controller is simple, so that I can come up with a single pattern that the controller starts from a state, reads the response, then generates a request and enters the next state. In many cases, the next state is even deterministic (I simplify the proof of termination for such steps).

Actually, I should not require 
```
forall |zk: ZookeeperClusterView, resp_o: Option<APIResponse>|
            #[trigger] reconcile_core(zk, resp_o, ZookeeperReconcileState{ reconcile_step: step_a }).0.reconcile_step == step_b
```
, and it is easy to make this more general so that the lemmas can apply to `step_a == AfterGetStatefulSet`.

Even when new operations is added to `reconcile_core` later, I believe that we can find similar behavior in different actions.

euclidgame:
But I really don't know why some lemmas take that long to verify ....

marshtompsxd:
> But I really don't know why some lemmas take that long to finish ....

Proof instability. Sometimes if you move the lemma to a different place or add an irrelevant lemma the verification time would change a lot. This is because the underlying z3 uses some seed to decide how to search for the proof and that seed is affected by the context. I think the Verus team is also working on that.

marshtompsxd:
I think the PR is no longer WIP?

marshtompsxd:
> and it is easy to make this more general so that the lemmas can apply to step_a == AfterGetStatefulSet.

The most general form is probably to make step_b a set of steps, but it is not super unclear to me how to derive the temporal predicate (a disjunction of the temporal predicates on each step in this set) from the set and how easy it would be to prove such a precondition.

Another solution is to have different forms of this lemma that handles steps that can lead to two or three different steps. This is acceptable because even if we further complicate the zk controller later, each step won't lead to too many different steps (usually three is enough).

But the current lemma is already very good, and I like it a lot.

euclidgame:
It is still WIP

PR #166, opened 2023-07-04
Simplify liveness proof of zookeeper controller

euclidgame:
In this PR, I mainly do two things:
 1. Some straightforward simplifications for liveness proof of zookeeper controller.
 2. Apply some lemmas (properties) of zookeeper controller to all controllers.

### DIFF
--- a/src/controller_examples/zookeeper_controller/proof/common.rs
+++ b/src/controller_examples/zookeeper_controller/proof/common.rs
@@ -42,50 +42,42 @@ pub open spec fn at_zookeeper_step_with_zk(zk: ZookeeperClusterView, step: Zooke
     }
 }
 
-pub open spec fn at_init_step_with_no_pending_req(zk: ZookeeperClusterView) -> StatePred<ClusterState> {
+pub open spec fn no_pending_req_at_zookeeper_step_with_zk(zk: ZookeeperClusterView, step: ZookeeperReconcileStep) -> StatePred<ClusterState> {
     |s: ClusterState| {
-        &&& at_zookeeper_step_with_zk(zk, ZookeeperReconcileStep::Init)(s)
+        &&& at_zookeeper_step_with_zk(zk, step)(s)
         &&& s.reconcile_state_of(zk.object_ref()).pending_req_msg.is_None()
     }
 }
 
-pub open spec fn is_create_headless_service_request_msg(msg: Message, zk: ZookeeperClusterView) -> bool {
-    &&& msg.src == HostId::CustomController
-    &&& msg.dst == HostId::KubernetesAPI
-    &&& msg.content.is_create_request()
-    &&& msg.content.get_create_request().namespace == zk.metadata.namespace.get_Some_0()
-    &&& msg.content.get_create_request().obj == make_headless_service(zk).to_dynamic_object()
-}
-
-pub open spec fn at_after_create_headless_service_step_with_zk_and_pending_req_in_flight(
-    zk: ZookeeperClusterView
+pub open spec fn pending_req_in_flight_at_zookeeper_step_with_zk(
+    step: ZookeeperReconcileStep, zk: ZookeeperClusterView, object: DynamicObjectView
 ) -> StatePred<ClusterState> {
     |s: ClusterState| {
-        &&& at_zookeeper_step_with_zk(zk, ZookeeperReconcileStep::AfterCreateHeadlessService)(s)
+        &&& at_zookeeper_step_with_zk(zk, step)(s)
         &&& s.reconcile_state_of(zk.object_ref()).pending_req_msg.is_Some()
         &&& s.message_in_flight(s.pending_req_of(zk.object_ref()))
-        &&& is_create_headless_service_request_msg(s.pending_req_of(zk.object_ref()), zk)
+        &&& is_correct_pending_request_at_zookeeper_step(step, s.pending_req_of(zk.object_ref()), zk, object)
     }
 }
 
-pub open spec fn req_msg_is_the_in_flight_pending_req_at_after_create_headless_service_step_with_zk(
-    zk: ZookeeperClusterView, req_msg: Message
+pub open spec fn req_msg_is_the_in_flight_pending_req_at_zookeeper_step_with_zk(
+    step: ZookeeperReconcileStep, zk: ZookeeperClusterView, req_msg: Message, object: DynamicObjectView
 ) -> StatePred<ClusterState> {
     |s: ClusterState| {
-        &&& at_zookeeper_step_with_zk(zk, ZookeeperReconcileStep::AfterCreateHeadlessService)(s)
+        &&& at_zookeeper_step_with_zk(zk, step)(s)
         &&& s.reconcile_state_of(zk.object_ref()).pending_req_msg == Option::Some(req_msg)
         &&& s.message_in_flight(req_msg)
-        &&& is_create_headless_service_request_msg(req_msg, zk)
+        &&& is_correct_pending_request_at_zookeeper_step(step, req_msg, zk, object)
     }
 }
 
-pub open spec fn at_after_create_headless_service_step_with_zk_and_exists_resp_in_flight(
-    zk: ZookeeperClusterView
+pub open spec fn exists_resp_in_flight_at_zookeeper_step_with_zk(
+    step: ZookeeperReconcileStep, zk: ZookeeperClusterView, object: DynamicObjectView
 ) -> StatePred<ClusterState> {
     |s: ClusterState| {
-        &&& at_zookeeper_step_with_zk(zk, ZookeeperReconcileStep::AfterCreateHeadlessService)(s)
+        &&& at_zookeeper_step_with_zk(zk, step)(s)
         &&& s.reconcile_state_of(zk.object_ref()).pending_req_msg.is_Some()
-        &&& is_create_headless_service_request_msg(s.pending_req_of(zk.object_ref()), zk)
+        &&& is_correct_pending_request_at_zookeeper_step(step, s.pending_req_of(zk.object_ref()), zk, object)
         &&& exists |resp_msg| {
             &&& #[trigger] s.message_in_flight(resp_msg)
             &&& resp_msg_matches_req_msg(resp_msg, s.pending_req_of(zk.object_ref()))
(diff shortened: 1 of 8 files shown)


src/controller_examples/zookeeper_controller/proof/common.rs
```diff
@@ -417,136 +157,84 @@ pub open spec fn at_after_get_stateful_set_step_with_zk_and_exists_not_found_err
...
 
-pub open spec fn at_after_create_stateful_set_step_with_zk_and_pending_req_in_flight(
-    zk: ZookeeperClusterView
-) -> StatePred<ClusterState> {
-    |s: ClusterState| {
-        &&& at_after_create_stateful_set_step_with_zk(zk)(s)
-        &&& s.reconcile_state_of(zk.object_ref()).pending_req_msg.is_Some()
-        &&& s.message_in_flight(s.pending_req_of(zk.object_ref()))
-        &&& is_create_stateful_set_request_msg(s.pending_req_of(zk.object_ref()), zk)
+pub open spec fn is_correct_pending_request_at_zookeeper_step(
```
marshtompsxd, line 169: Like it!

src/controller_examples/zookeeper_controller/proof/liveness.rs
```diff
@@ -1192,17 +897,113 @@ proof fn lemma_true_leads_to_always_current_state_matches_zk_under_eventual_inva
...
+        lift_state(pending_req_in_flight_at_zookeeper_step_with_zk(ZookeeperReconcileStep::AfterCreateClientService, zk, arbitrary())),
+        lift_state(pending_req_in_flight_at_zookeeper_step_with_zk(ZookeeperReconcileStep::AfterCreateAdminServerService, zk, arbitrary())),
+        lift_state(pending_req_in_flight_at_zookeeper_step_with_zk(ZookeeperReconcileStep::AfterCreateConfigMap, zk, arbitrary())),
+        lift_state(pending_req_in_flight_at_zookeeper_step_with_zk(ZookeeperReconcileStep::AfterGetStatefulSet, zk, arbitrary())),
         lift_state(current_state_matches(zk)),
         always(lift_state(current_state_matches(zk)))
     );
 }
 
+proof fn lemma_from_pending_req_in_flight_at_some_step_to_next_step(
```
marshtompsxd, line 911: This is nice!

marshtompsxd: I am even thinking about whether we can further generalize this lemma (make it applicable to all controllers), but let's leave it for future PRs

marshtompsxd: Could you document this lemma? Especially why we need these preconditions. The documentation will help us later a lot.

src/controller_examples/zookeeper_controller/proof/liveness.rs
```diff
@@ -1324,167 +1125,32 @@ proof fn lemma_from_init_step_to_after_create_headless_service_step(
...
-    }
-
-    kubernetes_api_liveness::lemma_pre_leads_to_post_by_kubernetes_api::<ZookeeperClusterView, ZookeeperReconcileState, ZookeeperReconciler>(
-        spec, input, stronger_next, handle_request(), pre, post
-    );
-}
-
-proof fn lemma_from_after_create_client_service_step_to_after_create_admin_server_service_step(
-    spec: TempPred<ClusterState>, zk: ZookeeperClusterView, resp_msg: Message
+proof fn lemma_from_resp_in_flight_at_some_step_to_pending_req_at_next_step(
```
marshtompsxd, line 1128: Similarly, could you document this lemma?

src/controller_examples/zookeeper_controller/proof/liveness.rs
```diff
@@ -1505,122 +1171,46 @@ proof fn lemma_from_after_create_client_service_step_to_after_create_admin_serve
...
-    );
-
-    assert forall |s, s_prime| pre(s) && #[trigger] stronger_next(s, s_prime) && kubernetes_api_next().forward(input)(s, s_prime)
-    implies post(s_prime) by {
-        let resp_msg = handle_create_request(req_msg, s.kubernetes_api_state).1;
-        assert({
-            &&& s_prime.message_in_flight(resp_msg)
-            &&& resp_msg_matches_req_msg(resp_msg, req_msg)
-        });
+    assert forall |s, s_prime| pre(s) && #[trigger] stronger_next(s, s_prime) implies pre(s_prime) || post(s_prime) by {
```
marshtompsxd, line 1174: Is this for proof stability?

src/controller_examples/zookeeper_controller/proof/liveness.rs
```diff
@@ -1192,17 +897,113 @@ proof fn lemma_true_leads_to_always_current_state_matches_zk_under_eventual_inva
...
+        spec.entails(always(lift_action(next::<ZookeeperClusterView, ZookeeperReconcileState, ZookeeperReconciler>()))),
+        spec.entails(tla_forall(|i| controller_next::<ZookeeperClusterView, ZookeeperReconcileState, ZookeeperReconciler>().weak_fairness(i))),
+        spec.entails(tla_forall(|i| kubernetes_api_next().weak_fairness(i))),
+        spec.entails(always(lift_state(crash_disabled()))),
+        spec.entails(always(lift_state(controller_runtime_safety::every_in_flight_msg_has_unique_id()))),
+        spec.entails(always(lift_state(controller_runtime_safety::each_resp_matches_at_most_one_pending_req(zk.object_ref())))),
+        step != ZookeeperReconcileStep::Error, step != ZookeeperReconcileStep::Done,
+        next_step != ZookeeperReconcileStep::Init,
+        forall |zk_1: ZookeeperClusterView, resp_o: Option<APIResponse>|
+            #[trigger] reconcile_core(zk_1, resp_o, ZookeeperReconcileState{ reconcile_step: step }).0.reconcile_step == next_step
```
marshtompsxd, line 924: I think this precondition wants to say no matter what, the next step is `next_step`. I am trying to understand why we have to forall zk_1 here, instead of just talking about zk.

src/controller_examples/zookeeper_controller/proof/liveness.rs
euclidgame: Yes, it is.

src/controller_examples/zookeeper_controller/proof/liveness.rs
euclidgame: We can't infer from the `pre` state that the cr passed to `reconcile_core` is exactly `zk`. However, we may weaken the precondition for this lemma.

src/controller_examples/zookeeper_controller/proof/liveness.rs
euclidgame: > I am even thinking about whether we can further generalize this lemma

This is what we need to do later.

euclidgame: And generalizing it seems not difficult but requires more handlings of "correct requests".

src/kubernetes_cluster/proof/controller_runtime_safety.rs
```diff
@@ -513,4 +514,90 @@ pub proof fn lemma_always_each_resp_matches_at_most_one_pending_req<K: ResourceV
...
 }
 
+// This lemma ensures that if a controller is at some reconcile state for a cr, there must be the pending request of the
+// reconcile state in flight or a correponding response in flight.
+// Obviously, this requires that when controller enters the 'state' in reconcile_core, there must be a request generated;
+// otherwise, the pending request may not be there.
+// The proof is very straightforward:
+//   - Right after the controller enters 'state', the pending request is added to in_flight.
+//   - If the pending request is processed by kubernetes api, there will be a response in flight.
+//   - If the response is processed by the controller, the controller will either enter a new state or a new pending request in flight.
```
marshtompsxd, line 524: > either enter a new state or a new pending request in flight

Or?

euclidgame: Yes. By this I mean the controller will not necessarily enter a new state; if it enters a state that is same as 'state', then a new pending request will be in flight, so in this case the invariant will not be violated.

marshtompsxd: I see what you mean. Maybe it is better to say "If the response is processed by the controller, the controller will either the next state (which could still be 'state' but the invariant still holds then)."

src/controller_examples/zookeeper_controller/proof/common.rs
```diff
@@ -130,252 +39,79 @@ pub open spec fn at_zookeeper_step(key: ObjectRef, step: ZookeeperReconcileStep)
...
     |s: ClusterState| {
-        &&& at_after_create_config_map_step_with_zk(zk)(s)
+        &&& at_zookeeper_step_with_zk(zk, step)(s)
         &&& s.reconcile_state_of(zk.object_ref()).pending_req_msg.is_Some()
-        &&& is_create_config_map_request_msg(s.pending_req_of(zk.object_ref()), zk)
+        &&& is_correct_pending_request_msg_at_zookeeper_step(step, s.pending_req_of(zk.object_ref()), zk, object)
         &&& s.message_in_flight(resp_msg)
         &&& resp_msg_matches_req_msg(resp_msg, s.pending_req_of(zk.object_ref()))
     }
 }
```
marshtompsxd, line 104: I think the above spec functions can be made more generic and will be reused heavily. But to do so we might need to make the `step` an associated type of `Reconciler` and the spec functions need to take a closure for `is_correct_pending_request_msg_at_zookeeper_step`. We can work on this later.

src/controller_examples/zookeeper_controller/proof/liveness.rs
```diff
@@ -1217,17 +922,129 @@ proof fn lemma_true_leads_to_always_current_state_matches_zk_under_eventual_inva
...
+        lift_state(pending_req_in_flight_at_zookeeper_step_with_zk(ZookeeperReconcileStep::AfterCreateClientService, zk, arbitrary())),
+        lift_state(pending_req_in_flight_at_zookeeper_step_with_zk(ZookeeperReconcileStep::AfterCreateAdminServerService, zk, arbitrary())),
+        lift_state(pending_req_in_flight_at_zookeeper_step_with_zk(ZookeeperReconcileStep::AfterCreateConfigMap, zk, arbitrary())),
+        lift_state(pending_req_in_flight_at_zookeeper_step_with_zk(ZookeeperReconcileStep::AfterGetStatefulSet, zk, arbitrary())),
         lift_state(current_state_matches(zk)),
         always(lift_state(current_state_matches(zk)))
     );
 }
 
+// This lemma ensures that zookeeper controller at some step with pending request in flight will finally enter its next step.
```
marshtompsxd, line 936: To avoid ambiguity, you can say "This lemma ensures that zookeeper controller at step a with pending request in flight will finally enter step b." and use step a, step b below, instead of initial, final step. My first reaction when seeing initial step is actually `ZookeeperReconcileStep::Init`

src/controller_examples/zookeeper_controller/proof/liveness.rs
```diff
@@ -1217,17 +922,129 @@ proof fn lemma_true_leads_to_always_current_state_matches_zk_under_eventual_inva
...
+// For the initial state and final state, they both require the reconcile_state to have a pending request which is the correct
+// request for that step (parameter 'step' for the initial step, parameter 'next_step' for the final state).
+// Note that in this lemma we add some constraints to the initial step:
+//    1. The next step of it when conducting reconcile_core is deterministic.
```
marshtompsxd, line 940: Instead of deterministic, I think what you mean is that there is only one possible next step (i.e., step b) for step a. The reconcile_core is still deterministic even with branches.

src/controller_examples/zookeeper_controller/proof/common.rs
euclidgame: Absolutely.

src/controller_examples/zookeeper_controller/proof/liveness.rs
```diff
@@ -1217,17 +922,129 @@ proof fn lemma_true_leads_to_always_current_state_matches_zk_under_eventual_inva
...
+//    2. When the controller enters this step, it must creates a request (which will be used to create the pending request message)
```
marshtompsxd, line 941: "which will be used to create the pending request message" --> "which is piggybacked by the pending request message"

Typo: creates -> create

src/controller_examples/zookeeper_controller/proof/common.rs
euclidgame: > take a closure for `is_correct_pending_request_msg_at_zookeeper_step`

Actually I have more thoughts about this... I will put it in the issue.

src/controller_examples/zookeeper_controller/proof/liveness.rs
```diff
@@ -1349,316 +1166,115 @@ proof fn lemma_from_init_step_to_after_create_headless_service_step(
...
-
-    kubernetes_api_liveness::lemma_pre_leads_to_post_by_kubernetes_api::<ZookeeperClusterView, ZookeeperReconcileState, ZookeeperReconciler>(
-        spec, input, stronger_next, handle_request(), pre, post
-    );
-}
-
-proof fn lemma_from_after_create_headless_service_step_to_after_create_client_service_step(
-    spec: TempPred<ClusterState>, zk: ZookeeperClusterView, resp_msg: Message
+// This lemma ensures that zookeeper controller at some step with a response in flight that matches its pending request will finally enter its next step.
+// For the initial state and final state, they both require the reconcile_state to have a pending request which is the correct
```
marshtompsxd, line 1170: Similar comment as above. Try to avoid using initial and final here

src/controller_examples/zookeeper_controller/proof/liveness.rs
```diff
@@ -1349,316 +1166,115 @@ proof fn lemma_from_init_step_to_after_create_headless_service_step(
...
+// request for that step (parameter 'step' for the initial step, parameter 'result_step' for the final state). For the initial state
+// alone, there is a known response (the parameter resp_msg) in flight that matches the pending request.
+// Note that in this lemma we add some constraints to the initial step:
+//    1. The next step of it when conducting reconcile_core is deterministic.
```
marshtompsxd, line 1174: Similar comment for "deterministic"

src/controller_examples/zookeeper_controller/proof/liveness.rs
```diff
@@ -1349,316 +1166,115 @@ proof fn lemma_from_init_step_to_after_create_headless_service_step(
...
+//    2. When the controller enters this step, it must creates a request (which will be used to create the pending request message)
```
marshtompsxd, line 1175: Similar comment

PR #167, opened 2023-07-05
Support k8s busy action and fix the liveness proof after that

euclidgame:
In this PR, we add a new action to the state machine that supports kubernetes to stay busy and reject the received request (send an error response).

### DIFF
--- a/src/kubernetes_api_objects/error.rs
+++ b/src/kubernetes_api_objects/error.rs
@@ -16,7 +16,7 @@ pub enum APIError {
     ObjectNotFound,
     ObjectAlreadyExists,
     NotSupported,
-    ServerBusy,
+    ServerTimeout,
     Other
 }
 
@@ -30,7 +30,7 @@ impl std::fmt::Debug for APIError {
             APIError::ObjectNotFound => write!(f, "ObjectNotFound"),
             APIError::ObjectAlreadyExists => write!(f, "ObjectAlreadyExists"),
             APIError::NotSupported => write!(f, "NotSupported"),
-            APIError::ServerBusy => write!(f, "ServerBusy"),
+            APIError::ServerTimeout => write!(f, "ServerTimeout"),
             APIError::Other => write!(f, "Other"),
         }
     }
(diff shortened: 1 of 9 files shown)


src/kubernetes_cluster/spec/cluster.rs
```diff
@@ -250,6 +252,50 @@ pub open spec fn disable_crash<K: ResourceView, T>() -> Action<State<K, T>, (),
     }
 }
 
+pub open spec fn make_kubernetes_api_busy<K: ResourceView, T>() -> Action<State<K, T>, Option<Message>, ()> {
```
marshtompsxd, line 255: Can you rename it to `kubernetes_api_is_busy`?

src/kubernetes_cluster/spec/cluster.rs
```diff
@@ -106,6 +107,7 @@ pub open spec fn init<K: ResourceView, T, ReconcilerType: Reconciler<K ,T>>() ->
         &&& (client::<K>().init)(s.client_state)
         &&& (network().init)(s.network_state)
         &&& s.crash_enabled
+        &&& s.k8s_maybe_busy
```
marshtompsxd, line 110: Can you rename it to `busy_enabled`?

src/kubernetes_cluster/spec/cluster.rs
euclidgame: Okay. I hesitated a lot on that and `kubernetes_api_is_busy` doesn't sound like an action.

marshtompsxd: That makes sense. How about `busy_kubernetes_api_rejects_request`

src/kubernetes_api_objects/error.rs
```diff
@@ -16,6 +16,7 @@ pub enum APIError {
     ObjectNotFound,
     ObjectAlreadyExists,
     NotSupported,
+    ServerBusy,
```
marshtompsxd, line 19: Can you rename it to `Timeout` or `ServerTimeout`? To align with the official doc: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#response-status-kind

src/kubernetes_cluster/spec/cluster.rs
euclidgame: `busy_kubernetes_api_rejects_request` sounds better but a bit long.

PR #169, opened 2023-07-06
Remove the next_state != init preconditions for some lemmas

euclidgame:
Previously, I add `next_state != init` to the preconditions of some lemmas for disabling the `run_scheduled_reconcile` action. Now I find this is really not needed because the state before `~>` requires that the reconcile already starts.


marshtompsxd:
I actually had the question that why we need that precondition but I didn't ask.

PR #177, opened 2023-07-14
Define and prove a naive deletion safety for zookeeper controller

euclidgame:
This is the very beginning of the safety property proof, and there are quite some factors to consider in terms of the specifications of the properties.

1. Should write the invariant as `forall |req_msg| sent_by_controller(req_msg) /\ in_flight(req_msg) => requirement(req_msg)` or  `!exists |req_msg| sent_by_controller(req_msg) /\ in_flight(req_msg) /\ !requirement(req_msg)`. Of course they are equivalent but are from different views.
2.  Different from the custom resources things, I think it's better to wrap the message inside a `tla_forall` (but I don't do it) instead of use a `forall` outside the `entails` predicate because the message is created during the execution.


src/controller_examples/zookeeper_controller/proof/liveness/liveness_property.rs
```diff
@@ -27,7 +27,7 @@ use crate::kubernetes_cluster::{
 use crate::temporal_logic::{defs::*, rules::*};
 use crate::zookeeper_controller::{
     common::*,
-    proof::{common::*, safety, terminate},
+    proof::{common::*, liveness::invariants as specific_invariants, liveness::terminate},
```
marshtompsxd, line 30: Maybe name invariants.rs to helper_invariants.rs?

euclidgame: Done.

Issue #174, opened 2023-07-09
Zookeeper scale down and scale up problems


ZichengMa:
#### Scale up

The scale up operation is mainly handled by a script called `zookeeperStart.sh` inside the zookeeper image. The script will check whether headless service is ready and an ensemble is already available or not. Besides, it checks whether there exists `Myid` file and `zoo.dynmaic.conf`, and these two files can tell is current pod a new member should be added in the ensemble or it's already a member but restarts for some reasons. If current pod should be added into the ensemble's peer list, the script will use a java file called `zu.jar` to reconfig the peer list, which is the same as using zkCli.sh to reconfig. 

#### Scale down

The scale up operation is mainly handled by a script called `zookeeperTeardown.sh` inside the zookeeper image. The script checks whether current pod should leave the ensemble permanently by comparing the `ClusterSize` and `MyId`. If `ClusterSize < MyId`, this server is being permanently removed. Otherwise, current pod is restarting due to some reasons. If current pod needs to be removed, the script calls `zu.jar` to reconfig.

#### Scale UP

The same as the official one. (i.e. the script handles the scale up)

#### Scale down

Suppose the new replica number is `replicas-new` and the old number is `replicas-old`

1. Send sts UPDATE request.    Let `sts.spec.replicas = replicas-new` 
2. Connect zkClient and get current peer list with `replicas-old` entries. Use slice from `0` to `replicas-new - 1` as new peer list
3. Send zkClient UPDATE_ENSEMBLE request to update peer list.

##### Can not delete to `quorum-1` at once

If `replicas-new` is smaller than quorum, which means `replicas-new <= (replicas-old + 1)/2`, and unfortunately the controller crashes after sending sts UPDATE, the zookeeper ensemble will directly down because there are not enough servers(at least quorum) working.

##### Can not delete more than 2 servers at once

[...]

PR #179, opened 2023-07-17
Change the way how unverified zookeeper controller scales down and scales up

ZichengMa:
Previously, we used a declartive way to update peer list of zookeeper cluster but we met some problems (#174) . After discussion, we decided to use the way how official controller handles scale down/up.


reference-controllers/zookeeper-controller/src/main.rs
```diff
@@ -217,48 +217,51 @@ async fn reconcile_stateful_set(zk: &ZookeeperCluster, client: Client) -> Result
...
+                            .to_vec(),
+                            Some(-1),
+                        )
+                        .map_err(Error::ClusterSizeZKNodeCreationFailed)?;
+                    info!(
+                        "Updated cluster size: {}",
+                        updated_sts.spec.as_ref().unwrap().replicas.unwrap()
+                    );
+                }
+                None => {}
```
marshtompsxd, line 258: It seems that we should not be silent if the node does not exist?

reference-controllers/zookeeper-controller/src/main.rs
```diff
@@ -217,48 +217,51 @@ async fn reconcile_stateful_set(zk: &ZookeeperCluster, client: Client) -> Result
...
-
-            // use ',' to join
-            let final_config = kept_servers.join(",");
-
-            info!("Modified config: {}", final_config);
-            let strings = vec![final_config.as_str()];
-            let test_update = EnsembleUpdate::New {
-                ensemble: strings.iter().cloned(),
+            let path = cluster_size_zk_node_path(zk);
+            info!("Try to create {} node since all replicas are ready", path);
```
marshtompsxd, line 227: Are we waiting for all replica pods to be ready?

reference-controllers/zookeeper-controller/src/main.rs
```diff
@@ -305,10 +308,7 @@ async fn reconcile_zk_node(zk: &ZookeeperCluster, client: Client) -> Result<(),
                 zk_client
                     .create(
                         "/zookeeper-operator",
-                        format!("CLUSTER_SIZE={}", zk.spec.replicas)
-                            .as_str()
-                            .as_bytes()
-                            .to_vec(),
+                        "".as_bytes().to_vec(),
```
marshtompsxd, line 311: Why not specify the cluster size when creating the node?

reference-controllers/zookeeper-controller/src/main.rs
```diff
@@ -217,48 +217,51 @@ async fn reconcile_stateful_set(zk: &ZookeeperCluster, client: Client) -> Result
             ..old_sts
         };
         // Scale up or no change
-        sts_api
-            .replace(sts_name, &PostParams::default(), &updated_sts)
-            .await
-            .map_err(Error::ReconcileStatefulSetFailed)?;
-
         if updated_sts.spec.as_ref().unwrap().replicas.unwrap()
             < old_sts.spec.as_ref().unwrap().replicas.unwrap()
```
marshtompsxd, line 221: It seems to me we don't need to check whether it is a scale down now? because updating the CLUSTER_SIZE does not harm to scale up.

reference-controllers/zookeeper-controller/src/main.rs
ZichengMa: Copy paste error

reference-controllers/zookeeper-controller/src/main.rs
ZichengMa: I think this data is meaningless. The official one leaves it as blank so I did so. The CLUSTER_SIZE is stored in the next layer(/zookeeper-operator/$zk-name)

PR #180, opened 2023-07-18
Support unverified third party APIs in controllers

euclidgame:
This PR allows developers to call any (unverified) third-party APIs besides the Kubernetes client APIs when implementing the controller.

It adds `Lib` as a generic type to `Reconciler`. `Lib` also takes two generic types `LibInputType` and `LibOutputType`. `Lib` wraps all interactions with the third-party APIs, and it implements the trait `ExternalLibrary<LibInputType, LibOutputType>`. `ExternalLibrary` has a method `process` which calls the third-party API in a synchronous way: it takes the input from `reconcile_core` (i.e., `LibInputType`) and returns an output (`Option<LibOutputType>`).

During every reconcile loop, the reconciler will return an `Option<Request>` (instead of the original `Option<KubeAPIRequest>`) to the `shim_layer`. `Request` is an enum with two variants for the two types of effects issued by the controller: invoking Kubernetes client APIs or third-party APIs. The shim layer will do pattern matching to the `Request` accordingly, which will either invoke the Kubernetes client API as before, or invoke the third-party APIs by calling the developer-provided `process`.

Correspondingly, the input to the reconciler, `Response`, is also an enum that encodes the two types of responses from the Kubernetes client APIs or third-party APIs.


src/reconciler/exec/io.rs
```diff
@@ -0,0 +1,89 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: MIT
+#![allow(unused_imports)]
+use crate::kubernetes_api_objects::{api_method::*, common::*};
+use crate::reconciler::spec::io::*;
+use vstd::{prelude::*, view::*};
+
+verus! {
+
+pub enum Receiver<T: View> {
```
marshtompsxd, line 10: Could you add more comments on what should be `T` here? Same for `Response`

src/reconciler/exec/external.rs
```diff
@@ -0,0 +1,28 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: MIT
+use vstd::{prelude::*, view::*};
+
+verus! {
+
+pub trait ExternalLibrary<InputType: View, OutputType: View> {
```
marshtompsxd, line 7: I think this is used when the controller does not need to communicate with any external systems? Could we have some doc (comments) here?

src/controller_examples/zookeeper_controller/common/mod.rs
```diff
@@ -19,4 +19,16 @@ pub enum ZookeeperReconcileStep {
     Error,
 }
 
+pub enum Error {
```
marshtompsxd, line 22: Do we use these errors anywhere?

src/controller_examples/zookeeper_controller/exec/zookeeper_lib/helper_funcs.rs
```diff
@@ -0,0 +1,76 @@
...
+
+verus! {
+
+struct LoggingWatcher;
+impl Watcher for LoggingWatcher {
+    #[verifier(external)]
+    fn handle(&self, e: WatchedEvent) {}
+}
+
+pub fn empty_func(zk: String) -> Result<(), Error> {
```
marshtompsxd, line 18: I might miss sth, but what is this empty_func used for?

src/controller_examples/zookeeper_controller/exec/zookeeper_lib/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+impl View for ZKSupportOutput {
+    type V = ZKSupportOutputView;
+    spec fn view(&self) -> ZKSupportOutputView {
+        match self {
+            ZKSupportOutput::ReconcileZKNode(res) => ZKSupportOutputView::ReconcileZKNode(*res),
+        }
+    }
+}
+
+// pub open spec fn
```
marshtompsxd, line 38: nit: let's remove this line

src/reconciler/exec/io.rs
```diff
@@ -0,0 +1,101 @@
...
+// if it want the third-party library to deal with the request, it should return a Request::ExternalRequest.
+pub enum Request<T: View> {
+    KRequest(KubeAPIRequest),
+    ExternalRequest(T),
+}
+
+// The response type should be consistent with the Request type.
+// T: the output type of the third-party library of the reconciler which should be defined by the developer.
+// A safe and common way is to have an enum type which has the corresponding types of the input type in the Request.
+// Anway, the process method in the ExternalLibrary, the input type in Request, output type in Response and the handling
```
marshtompsxd, line 26: Anyway?

src/reconciler/exec/external.rs
```diff
@@ -0,0 +1,36 @@
...
+
+verus! {
+
+// A trait for the external library of a reconciler.
+// Its core is a process method, and the developer should wrap all possible operations they may need in the function.
+// The InputType is the ? of Request<?> of the reconciler, i.e., it completes the request type of a reconciler.
+// Similarly, the OutputType composes the Response<?> type of a reconciler.
+// Note that we can encapsulate all the required libraries here, so each reconciler only has one ExternalLibrary type.
+pub trait ExternalLibrary<InputType: View, OutputType: View> {
+    fn process(input: InputType) -> Option<OutputType>;
```
marshtompsxd, line 13: Sometimes the third party APIs are async functions. Since process is not async, we probably should check whether Verus supports async block (in external functions) so that we can call async functions in process

src/kubernetes_cluster/mod.rs
```diff
@@ -1,4 +1,4 @@
 // Copyright 2022 VMware, Inc.
 // SPDX-License-Identifier: MIT
-pub mod proof;
+// pub mod proof;
```
marshtompsxd, line 3: Do you want to fix the proof lemmas in this PR or later?

src/reconciler/exec/io.rs
euclidgame: What should it be? I mean, although I introduce concretely what `T` is for, it is up to the developers to define it together with the Lib, reconcile_core process.

src/reconciler/exec/external.rs
euclidgame: What does that mean? I think we have already used some asyn blocks.

src/kubernetes_cluster/mod.rs
euclidgame: Later.

src/reconciler/exec/io.rs
marshtompsxd: oh I was just saying there is a typo

src/reconciler/exec/external.rs
marshtompsxd: You mean the one in `run_controller`
```
    let reconcile = |cr: Arc<K>, ctx: Arc<Data>| async move {
        return reconcile_with::<K, ResourceWrapperType, ReconcilerType, ReconcileStateType>(
            &ReconcilerType::default(), cr, ctx
        ).await;
    };
```
?

marshtompsxd: It is an async closure but not an async block, tho they are very similar

src/reconciler/exec/io.rs
```diff
@@ -0,0 +1,101 @@
...
+
+pub open spec fn opt_response_to_view<T: View>(resp: &Option<Response<T>>) -> Option<ResponseView<T::V>> {
+    match resp {
+        Option::Some(resp) => Option::Some(resp.to_view()),
+        Option::None => Option::None,
+    }
+}
+
+pub open spec fn opt_request_to_view<T: View>(receiver: &Option<Request<T>>) -> Option<RequestView<T::V>> {
+    match receiver {
```
marshtompsxd, line 95: Can you change receiver to request here?

marshtompsxd:
@euclidgame I also revised the description a bit. Please take a look to ensure it is still correct.

PR #181, opened 2023-07-18
Support config reload for rabbitmq unverified controller

ZichengMa:
Currently, timestamp are used to indicate whether the statefulset needs restarting. The ResourceVersion may be used as new indicator in next PR.

### DIFF
--- a/reference-controllers/rabbitmq-controller/src/main.rs
+++ b/reference-controllers/rabbitmq-controller/src/main.rs
@@ -20,7 +20,7 @@ use k8s_openapi::api::core::v1 as corev1;
 use k8s_openapi::api::rbac::v1 as rbacv1;
 use k8s_openapi::api::{apps::v1 as appsv1, apps::v1::StatefulSet, core::v1::Service};
 use kube::{
-    api::{Api, ListParams, PostParams},
+    api::{Api, ApiResource, DynamicObject, GroupVersionKind, ListParams, PostParams},
     client::ConfigExt,
     runtime::controller::{Action, Controller},
     Client, Config, CustomResourceExt,
@@ -54,6 +54,9 @@ enum Error {
 
     #[error("Failed to reconcile StatefulSet: {0}")]
     ReconcileStatefulSetFailed(#[source] kube::Error),
+
+    #[error("Failed to update annotation: {0}")]
+    AnnotationUpdateFailed(#[source] kube::Error),
 }
 
 struct Data {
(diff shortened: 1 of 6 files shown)


reference-controllers/rabbitmq-controller/src/main.rs
```diff
@@ -107,10 +142,171 @@ async fn reconcile_stateful_set(rabbitmq: &RabbitmqCluster, client: Client) -> R
             .create(&PostParams::default(), &sts)
             .await
             .map_err(Error::ReconcileStatefulSetFailed)?;
+        let gvk = GroupVersionKind::gvk("apps", "v1", "StatefulSet");
+        let api_resource = ApiResource::from_gvk(&gvk);
+        update_annotation(
```
marshtompsxd, line 147: Why do we need this "createdAt" annotation?

reference-controllers/rabbitmq-controller/src/main.rs
```diff
@@ -107,10 +142,171 @@ async fn reconcile_stateful_set(rabbitmq: &RabbitmqCluster, client: Client) -> R
...
+                    server_config_name,
+                    &PostParams::default(),
+                    &updated_server_config,
+                )
+                .await
+                .map_err(Error::ReconcileConfigMapFailed)?;
+
+            let gvk = GroupVersionKind::gvk("", "v1", "ConfigMap");
+            let api_resource = ApiResource::from_gvk(&gvk);
+            update_annotation(
```
marshtompsxd, line 211: Why do we need the annotation for configmap?

reference-controllers/rabbitmq-controller/src/main.rs
```diff
@@ -107,10 +142,171 @@ async fn reconcile_stateful_set(rabbitmq: &RabbitmqCluster, client: Client) -> R
...
+        return Ok(());
+    }
+
+    // Restart the sts by changing the pod template.
+    let patch = Patch::Merge(json!({
+        "spec": {
+            "template": {
+                "metadata": {
+                    "annotations": {
+                        "restartedAt": Utc::now().to_rfc3339(),
```
marshtompsxd, line 293: Can you try using configmap's rv, instead of timestamp, to restart the pods?

reference-controllers/rabbitmq-controller/src/main.rs
ZichengMa: We don't need that. It is for enabling feature flags(this is a feature that enable plugins' features, but it is not an important feature). I will delete it and if we want to implement this feature, I can add it back.

reference-controllers/rabbitmq-controller/src/main.rs
ZichengMa: It is used to store when did the configmap newly updated. We can delete it when we use resourceVerison(maybe next PR)

PR #182, opened 2023-07-19
Use ResourceVersion as the indicator for config reloading

ZichengMa:
We now use ResourceVersion to reload config.
When server-conf configmap is updated, it will annotate `lastVersion` as the same as its resourceVersion. The `lastVersion` is used to indicate configmap has been updated, but it will not be used to compare with sts's version.
When the sts is restarted, it will be annotated with `restartedVersion` == configmap's current resourceVersion.
In the sts restart processes, controller will check if sts `restartedVersion` is smaller than configmap's current resourceVersion. If the condition holds, which means configmap is updaed after last time the sts restarted, controller will restart sts by updating the podtemplate's anntation to current configmap resourceVersion.


reference-controllers/rabbitmq-controller/src/main.rs
```diff
@@ -201,8 +188,8 @@ async fn reconcile_server_config_map(
                 api_resource,
                 rabbitmq.metadata.namespace.as_ref().unwrap(),
                 &server_config_name,
-                "updatedAt",
-                &Utc::now().to_rfc3339(),
+                "lastVersion",
+                &resrouce_version,
```
marshtompsxd, line 192: typo

reference-controllers/rabbitmq-controller/src/main.rs
```diff
@@ -201,8 +188,8 @@ async fn reconcile_server_config_map(
                 api_resource,
                 rabbitmq.metadata.namespace.as_ref().unwrap(),
                 &server_config_name,
-                "updatedAt",
-                &Utc::now().to_rfc3339(),
+                "lastVersion",
```
marshtompsxd, line 191: Can you rename this to lastConfigMapVersion

marshtompsxd: I just realized that this is in `reconcile_server_config_map` -- why do we need to update the configmap's annotation?

reference-controllers/rabbitmq-controller/src/main.rs
```diff
@@ -262,12 +250,18 @@ async fn restart_sts_if_needed(rabbitmq: &RabbitmqCluster, client: &Client) -> R
...
         None => None,
     };
 
+    let server_config_updatedat = server_config_o
+        .as_ref()
+        .unwrap()
+        .metadata
+        .resource_version
+        .clone();
     if sts_restartedat.is_some()
```
marshtompsxd, line 263: I think we should not call it `sts_restartedat` anymore?

reference-controllers/rabbitmq-controller/src/main.rs
```diff
@@ -262,12 +250,18 @@ async fn restart_sts_if_needed(rabbitmq: &RabbitmqCluster, client: &Client) -> R
...
-        && sts_restartedat.clone().unwrap() > server_config_updatedat.clone().unwrap()
+        && sts_restartedat.clone().unwrap() >= server_config_updatedat.clone().unwrap()
```
marshtompsxd, line 264: I don't think you have to check whether the version has changed -- if you update the statefulset with the exactly same annotation, there won't be any change and pods should not be restarted. But pls do some tests to validate what I said.

reference-controllers/rabbitmq-controller/src/main.rs
```diff
@@ -237,49 +237,16 @@ async fn restart_sts_if_needed(rabbitmq: &RabbitmqCluster, client: &Client) -> R
     );
```
marshtompsxd, line 237: Do we need a separate `restart_sts_if_needed` or we can just merge it into the branch that updates the sts inside `reconcile_stateful_set`?

marshtompsxd:
@ZichengMa please ping me again after you addressed what we discussed this afternoon

reference-controllers/rabbitmq-controller/src/main.rs
```diff
@@ -163,138 +140,88 @@ async fn reconcile_server_config_map(
...
+        let resource_version = server_config_o
+            .as_ref()
+            .unwrap()
+            .metadata
+            .resource_version
+            .as_ref()
+            .unwrap()
+            .clone();
+        info!("Current rv of server configmap: {}", resource_version);
         if server_config.data != server_config_o.as_ref().unwrap().data {
```
marshtompsxd, line 152: I would say we can just go ahead to update the configmap without checking whether the data is the same or not -- if the data remains the same then the update won't change anything.

reference-controllers/rabbitmq-controller/src/statefulset.rs
```diff
@@ -237,6 +231,10 @@ fn pod_template_spec(rabbitmq: &RabbitmqCluster) -> corev1::PodTemplateSpec{
                 "app".to_string(),
                 rabbitmq.meta().name.as_ref().unwrap().clone(),
             )])),
+            annotations: Some(BTreeMap::from([(
+                "restartedVersion".to_string(),
```
marshtompsxd, line 235: Can we rename it to "configMapVersionInUse"?

PR #183, opened 2023-07-20
Add reconcile_zk_node to the reconcile process of zookeeper controller

euclidgame:


marshtompsxd:
Is this PR still work in progress?

ZichengMa (approved):

marshtompsxd:
And should I review it?

ZichengMa:
> And should I review it?

Yes, I think so. I checked what Wenjie was doing and understanded the workflow. You can review it too.

euclidgame:
Not work in progress @marshtompsxd

src/controller_examples/zookeeper_controller/exec/zookeeper_lib/helper_funcs.rs
```diff
@@ -9,62 +9,65 @@ use vstd::{prelude::*, string::*};
...
+        Some(_) => {
+            return ZKNodeResult{ res: Ok(())};
+        },
         None => {
             // First create the parent node
             if zk_client
                 .create(
                     "/zookeeper-operator",
-                    new_strlit("CLUSTER_SIZE=").to_string().concat(i32_to_string(zk.spec().replicas()).as_str())
+                    new_strlit("CLUSTER_SIZE=").to_string().concat(replicas.as_str())
```
marshtompsxd, line 47: Why do we need to write the CLUSTER_SIZE to the parent node? I thought that only the value in the child node matters (to the teardown script). Please correct me if I am wrong @ZichengMa @euclidgame

src/controller_examples/zookeeper_controller/exec/zookeeper_lib/helper_funcs.rs
```diff
@@ -9,62 +9,65 @@ use vstd::{prelude::*, string::*};
...
     let path_look_up = zk_client.exists(path.as_rust_string_ref(), false);
     if path_look_up.is_err() {
-        return Result::Err(Error::ClusterSizeZKNodeCreationFailed);
+        return ZKNodeResult{ res: Err(Error::ClusterSizeZKNodeCreationFailed) };
     }
     let path_look_up = path_look_up.unwrap();
     match path_look_up {
-        Some(_) => Ok(()),
+        Some(_) => {
+            return ZKNodeResult{ res: Ok(())};
```
marshtompsxd, line 40: Does it mean that we will never update the `CLUSTER_SIZE` once the zk node gets created? Then how to support downscaling?

src/controller_examples/zookeeper_controller/exec/reconciler.rs
```diff
@@ -214,17 +214,59 @@ pub fn reconcile_core(
         },
         ZookeeperReconcileStep::AfterCreateStatefulSet => {
             let state_prime = ZookeeperReconcileState {
-                reconcile_step: ZookeeperReconcileStep::Done,
+                reconcile_step: ZookeeperReconcileStep::AfterCreateZKNode,
                 ..state
             };
-            return (state_prime, Option::None);
+            let s1 = cluster_size_zk_node_path(zk);
```
marshtompsxd, line 220: Let's try to find better names than s1,2,3 :-))
I am totally fine if it takes more lines

src/controller_examples/zookeeper_controller/exec/zookeeper_lib/helper_funcs.rs
euclidgame: Oh I'm idiot on this, all I do is to follow the reference controller. Can you check the reference controller?

src/controller_examples/zookeeper_controller/exec/reconciler.rs
euclidgame: Sure.

src/controller_examples/zookeeper_controller/exec/zookeeper_lib/helper_funcs.rs
marshtompsxd: Confirmed that we don't need to set the value to the parent node.

src/controller_examples/zookeeper_controller/exec/zookeeper_lib/helper_funcs.rs
marshtompsxd: @euclidgame @ZichengMa Can you address this one?

ZichengMa: I will later add the Update operation and get familar with how to import third party APIs.

src/controller_examples/zookeeper_controller/exec/zookeeper_lib/lib.rs
```diff
@@ -1,36 +1,70 @@
...
 }
 
-impl View for ZKSupportInput {
+impl ToView for ZKSupportInput {
     type V = ZKSupportInputView;
-    spec fn view(&self) -> ZKSupportInputView {
+    spec fn to_view(&self) -> ZKSupportInputView {
         match self {
-            ZKSupportInput::ReconcileZKNode(zk) => ZKSupportInputView::ReconcileZKNode(zk@),
+            ZKSupportInput::ReconcileZKNode(s1,s2,s3) => ZKSupportInputView::ReconcileZKNode(s1@,s2@,s3@),
```
ZichengMa, line 27: s1,s2,s3 are not changed yet?

ZichengMa (changes requested):

src/controller_examples/zookeeper_controller/exec/zookeeper_lib/lib.rs
```diff
@@ -41,8 +77,8 @@ impl ExternalLibrary<ZKSupportInput, ZKSupportOutput> for ZKSupport {
     #[verifier(external)]
     fn process(input: ZKSupportInput) -> Option<ZKSupportOutput> {
         match input {
-            ZKSupportInput::ReconcileZKNode(zk)
-                => Option::Some(ZKSupportOutput::ReconcileZKNode(helper_funcs::reconcile_zk_node(&zk))),
+            ZKSupportInput::ReconcileZKNode(s1,s2,s3)
```
marshtompsxd, line 80: Can you give them more descriptive names as @ZichengMa suggested then I will merge the PR

marshtompsxd:
Thinking twice, we can further simplify the zk helper func because there is no need to include the zk's name into the path. Maybe @ZichengMa can validate whether the teardown script can still work well in that way. Let's worry about that in some next PR.

PR #185, opened 2023-07-22
Support external APIs in the state machine

euclidgame:
The core idea is that we add execution of external functions to the `continue_reconcile` action. 

To support this, we add `lib_response` to the controller state struct. Then there are three types of actions in `continue_reconcile` which are categorized based on the return request of `reconcile_core`:
1. If the `reconcile_core` returns a k8s api request, then we set the `pending_req_msg` field as before;
2. If the `reconcile_core` returns a external request, then we directly call the external library to process the request, which means we make the two steps done in one action. Then we set the `lib_response` field to the returned response.
3. If the `reconcile_core` returns no request, just leave both of the fields `Option::None`.

We should be careful about the preconditions of `continue_reconcile`. If the reconcile state has a pending request message, then the `lib_response` field must be empty because this means for this round of reconcile, we only process the k8s api response.

### DIFF
--- a/src/controller_examples/rabbitmq_controller/proof/common.rs
+++ b/src/controller_examples/rabbitmq_controller/proof/common.rs
@@ -18,10 +18,10 @@ use vstd::prelude::*;
 
 verus! {
 
-pub type ClusterState = State<RabbitmqClusterView, RabbitmqReconcileState>;
+pub type ClusterState = State<RabbitmqClusterView, RabbitmqReconciler>;
 
 pub open spec fn cluster_spec() -> TempPred<ClusterState> {
-    sm_spec::<RabbitmqClusterView, RabbitmqReconcileState, RabbitmqReconciler>()
+    sm_spec::<RabbitmqClusterView, RabbitmqReconciler>()
 }
 
 pub open spec fn rabbitmq_reconcile_state(step: RabbitmqReconcileStep) -> RabbitmqReconcileState {
@@ -52,7 +52,7 @@ pub open spec fn at_rabbitmq_step_with_rabbitmq(rabbitmq: RabbitmqClusterView, s
 pub open spec fn no_pending_req_at_rabbitmq_step_with_rabbitmq(rabbitmq: RabbitmqClusterView, step: RabbitmqReconcileStep) -> StatePred<ClusterState> {
     |s: ClusterState| {
         &&& at_rabbitmq_step_with_rabbitmq(rabbitmq, step)(s)
-        &&& no_pending_request(s, rabbitmq.object_ref)
+        &&& no_pending_request(s, rabbitmq.object_ref())
     }
 }
 
--- a/src/controller_examples/rabbitmq_controller/proof/common.rs
+++ b/src/controller_examples/rabbitmq_controller/proof/common.rs
@@ -4,6 +4,7 @@
 use crate::kubernetes_api_objects::{
     api_method::*, common::*, config_map::*, dynamic::*, resource::*, stateful_set::*,
 };
+use crate::kubernetes_cluster::proof::controller_runtime::*;
 use crate::kubernetes_cluster::spec::{
     cluster::*,
     controller::common::{controller_req_msg, ControllerActionInput, ControllerStep},
@@ -51,7 +52,7 @@ pub open spec fn at_rabbitmq_step_with_rabbitmq(rabbitmq: RabbitmqClusterView, s
 pub open spec fn no_pending_req_at_rabbitmq_step_with_rabbitmq(rabbitmq: RabbitmqClusterView, step: RabbitmqReconcileStep) -> StatePred<ClusterState> {
     |s: ClusterState| {
         &&& at_rabbitmq_step_with_rabbitmq(rabbitmq, step)(s)
-        &&& s.reconcile_state_of(rabbitmq.object_ref()).pending_req_msg.is_None()
+        &&& no_pending_request(s, rabbitmq.object_ref)
     }
 }
 
@@ -60,7 +61,7 @@ pub open spec fn pending_req_in_flight_at_rabbitmq_step_with_rabbitmq(
 ) -> StatePred<ClusterState> {
     |s: ClusterState| {
         &&& at_rabbitmq_step_with_rabbitmq(rabbitmq, step)(s)
-        &&& s.reconcile_state_of(rabbitmq.object_ref()).pending_req_msg.is_Some()
+        &&& pending_k8s_api_req_msg(s, rabbitmq.object_ref())
         &&& s.message_in_flight(s.pending_req_of(rabbitmq.object_ref()))
         &&& is_correct_pending_request_msg_at_rabbitmq_step(step, s.pending_req_of(rabbitmq.object_ref()), rabbitmq, object)
     }
@@ -71,7 +72,7 @@ pub open spec fn req_msg_is_the_in_flight_pending_req_at_rabbitmq_step_with_rabb
 ) -> StatePred<ClusterState> {
     |s: ClusterState| {
         &&& at_rabbitmq_step_with_rabbitmq(rabbitmq, step)(s)
-        &&& s.reconcile_state_of(rabbitmq.object_ref()).pending_req_msg == Option::Some(req_msg)
+        &&& pending_k8s_api_req_msg_is(s, rabbitmq.object_ref(), req_msg)
         &&& s.message_in_flight(req_msg)
         &&& is_correct_pending_request_msg_at_rabbitmq_step(step, req_msg, rabbitmq, object)
     }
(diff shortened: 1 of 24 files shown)


src/kubernetes_cluster/spec/controller/common.rs
```diff
@@ -17,7 +17,14 @@ pub struct ControllerState<K: ResourceView, R: Reconciler<K>> {
...
 pub struct OngoingReconcile<K: ResourceView, R: Reconciler<K>> {
     pub triggering_cr: K,
+    // Use two fields for pending k8s api request and pending external library result separately.
+    // Such design is not very reasonable because it doesn't make it apparent that the two fields
+    // are mutually exclusive, i.e., only one of them can be non-empty at some point.
+    // But this is fine, since we can describe the state as one of them is Option::Some(...) and
+    // the other is Option::None.
     pub pending_req_msg: Option<Message>,
+    pub pending_lib_req: Option<R::I>,
+    pub lib_response: Option<R::O>,
```
marshtompsxd, line 27: Do we plan to put "controller issues a pending lib req" and "controller receives the lib response" in the same action? If so we probably don't need `pending_lib_req` as part of the state?

src/controller_examples/rabbitmq_controller/proof/common.rs
```diff
@@ -17,10 +18,10 @@ use vstd::prelude::*;
 
 verus! {
 
-pub type ClusterState = State<RabbitmqClusterView, RabbitmqReconcileState>;
+pub type ClusterState = State<RabbitmqClusterView, RabbitmqReconciler>;
```
marshtompsxd, line 21: Is it true that all the changes inside rabbitmq_controller::proof are this type of refactoring?

src/kubernetes_cluster/spec/controller/controller_runtime.rs
```diff
@@ -48,52 +50,110 @@ pub open spec fn continue_reconcile<K: ResourceView, R: Reconciler<K>>() -> Cont
                 &&& s.ongoing_reconciles.dom().contains(cr_key)
                 &&& !R::reconcile_done(s.ongoing_reconciles[cr_key].local_state)
                 &&& !R::reconcile_error(s.ongoing_reconciles[cr_key].local_state)
-                &&& if input.recv.is_Some() {
+                // Split cases:
+                // (1) there is a pending request which is destined for kubernetes api;
+                // (2) there is a pending request which is destined for external library and no response;
+                // (3) there is a pending request which is destined for external library and a corresponding response;
+                // (4) there is no pending request;
```
marshtompsxd, line 57: Since we don't plan to have a separate state machine for external lib, I am wondering whether we want to split (2) and (3). An alternative is that in the same action, if the controller issues an external lib call, we run the "process" function and get the return value of the external lib and then store it in `lib_response`. Of course, this model has some limitations in its expressiveness, but it is better to start from a very simple one.

src/kubernetes_cluster/spec/controller/common.rs
euclidgame: Then how to make the exec code match the spec code?

marshtompsxd: The spec `reconcile_core` still returns lib req (or k8s req), but you can "process" it and get the lib response in the same action (basically, make the interaction with external lib atomic). And you will need a spec version of "process".

src/controller_examples/rabbitmq_controller/proof/common.rs
euclidgame: What do you mean?

marshtompsxd: I mean, I see you changed some type name and function name. Are all of these changes in the proof for such refactoring? I just want to see how careful I should be when going through these changes.

euclidgame: They are not all of the same type. And if the refactoring is wrong, the compiler will complain.

marshtompsxd:
`error: function body check: Resource limit (rlimit) exceeded; consider rerunning with --profile for more details`

What would happen if you run profile on it? And does the verification pass on your local machine? If it requires more resources maybe we should allocate more

src/kubernetes_cluster/spec/controller/controller_runtime.rs
```diff
@@ -48,52 +49,86 @@ pub open spec fn continue_reconcile<K: ResourceView, R: Reconciler<K>>() -> Cont
...
+                // the recv field of input is empty.
+                &&& if s.ongoing_reconciles[cr_key].pending_req_msg.is_Some() {
+                    &&& s.ongoing_reconciles[cr_key].lib_response.is_None() // The response from external library should have been consumed if ever exists
+                    &&& input.recv.is_Some()
                     &&& input.recv.get_Some_0().content.is_APIResponse()
-                    &&& s.ongoing_reconciles[cr_key].pending_req_msg.is_Some()
                     &&& resp_msg_matches_req_msg(input.recv.get_Some_0(), s.ongoing_reconciles[cr_key].pending_req_msg.get_Some_0())
                 } else {
-                    s.ongoing_reconciles[cr_key].pending_req_msg.is_None()
+                    // We should not get response from other part of the GSM for this case.
```
marshtompsxd, line 67: What is GSM? Global state machine?

euclidgame:
> `error: function body check: Resource limit (rlimit) exceeded; consider rerunning with --profile for more details`
> 
> What would happen if you run profile on it? And does the verification pass on your local machine? If it requires more resources maybe we should allocate more

This is quite weird. On my own laptop, this function also exceeds the time limit. But if I only verify the single function, it can pass quickly.

src/kubernetes_cluster/spec/controller/controller_runtime.rs
euclidgame: Yes

src/kubernetes_cluster/spec/controller/controller_runtime.rs
```diff
@@ -48,52 +49,86 @@ pub open spec fn continue_reconcile<K: ResourceView, R: Reconciler<K>>() -> Cont
...
-            let s_prime = ControllerState {
-                ongoing_reconciles: s.ongoing_reconciles.insert(cr_key, reconcile_state_prime),
-                ..s
-            };
-            let send = if pending_req_msg.is_Some() {
-                Multiset::singleton(pending_req_msg.get_Some_0())
+                        // Then update the state.
+                        let reconcile_state_prime = OngoingReconcile {
+                            pending_req_msg: Option::None,
+                            lib_response: Option::None, // TODO: update this field after add the library as a generic
```
marshtompsxd, line 108: Do you want to leave it for the next PR of finish it in this one?

euclidgame: Either work. But I must first let the verification go through before adding one more generic to the `Reconciler` trait.

marshtompsxd:
> > `error: function body check: Resource limit (rlimit) exceeded; consider rerunning with --profile for more details`
> > What would happen if you run profile on it? And does the verification pass on your local machine? If it requires more resources maybe we should allocate more
> 
> This is quite weird. On my own laptop, this function also exceeds the time limit. But if I only verify the single function, it can pass quickly.

Are you able to enlarge the time limit? (See if verus --help has such options) Another option is to divide that lemma into several smaller ones.

One more question: is the total verification time significantly larger than before this PR?

src/kubernetes_cluster/spec/controller/controller_runtime.rs
marshtompsxd: > But I must first let the verification go through before adding one more generic to the Reconciler trait.

Agree

src/reconciler/spec/reconciler.rs
```diff
@@ -39,6 +39,10 @@ pub trait Reconciler<#[verifier(maybe_negative)] K: ResourceView>: Sized {
     // reconcile_error is used to tell the controller_runtime whether this reconcile round returns with error.
     // If it is true, controller_runtime will requeue the reconcile.
     open spec fn reconcile_error(state: Self::T) -> bool;
+
+    // To avoid the super annoying refactoring, I first add an interface here for the processing of external library
+    // rather than add a generic to this trait.
```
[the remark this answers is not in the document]
marshtompsxd: If you want to leave some notes to remind yourself (in the future), I would recommend `TODO(Wenjie): xxxx`.

src/kubernetes_cluster/spec/controller/controller_runtime.rs
```diff
@@ -100,12 +100,10 @@ pub open spec fn continue_reconcile<K: ResourceView, R: Reconciler<K>>() -> Cont
...
                     },
                     RequestView::ExternalRequest(req) => {
-                        // Call the external library first.
-
-                        // Then update the state.
+                        // Update the state by calling the external process method.
                         let reconcile_state_prime = OngoingReconcile {
                             pending_req_msg: Option::None,
-                            lib_response: Option::None, // TODO: update this field after add the library as a generic
+                            lib_response: R::external_process(req), // TODO: update this field after add the library as a generic
```
marshtompsxd, line 106: Seems that this TODO is already done?

marshtompsxd:
Great work!

PR #186, opened 2023-07-24
Support scale down by updating zk node in the verified zk controller

ZichengMa:
The key is to update the `CLUSTER_SIZE` zk node to make it consistent with the expected replicas number so that the (unverified) teardown script inside the zookeeper image can correctly decide whether to remove the pod from the zookeeper cluster before the pod deletion.

To implement this, this PR also introduces some miscellaneous changes including (1) introducing one more field to `ZookeeperReconcileState` to store a statefulset object from the get request, (2) implementing the update branch of the external `reconcile_zk_node` function, (3) adding two more fields to the statefulset spec.


src/kubernetes_api_objects/persistent_volume_claim.rs
```diff
@@ -313,4 +313,98 @@ impl Marshalable for PersistentVolumeClaimSpecView {
...
+#[verifier(external_body)]
+pub struct StatefulSetPersistentVolumeClaimRetentionPolicy {
+    inner: deps_hack::k8s_openapi::api::apps::v1::StatefulSetPersistentVolumeClaimRetentionPolicy,
+}
+
+impl StatefulSetPersistentVolumeClaimRetentionPolicy {
+    pub spec fn view(&self) -> StatefulSetPersistentVolumeClaimRetentionPolicyView;
+
+    #[verifier(external_body)]
+    pub fn default() -> (pvc_spec: StatefulSetPersistentVolumeClaimRetentionPolicy)
```
marshtompsxd, line 325: We should not name this variable as pvc_spec

src/kubernetes_api_objects/pod.rs
```diff
@@ -500,6 +508,80 @@ impl Volume {
...
+        );
+    }
+
+    #[verifier(external)]
+    pub fn into_kube(self) -> deps_hack::k8s_openapi::api::core::v1::LifecycleHandler {
+        self.inner
+    }
+}
+
+
```
marshtompsxd, line 579: Can you remove the redundant empty lines?

src/kubernetes_api_objects/pod.rs
```diff
@@ -1081,6 +1172,46 @@ impl Marshalable for ContainerView {
...
+pub struct LifecycleHandlerView {
+    pub execs: Option<Seq<StringView>>,
+}
+
+impl LifecycleHandlerView {
+    pub open spec fn default() -> LifecycleHandlerView {
+        LifecycleHandlerView {
+            execs: Option::None,
```
marshtompsxd, line 1203: This should be `exec`, instead of `execs`, I guess? Please double check

src/kubernetes_api_objects/pod.rs
```diff
@@ -500,6 +508,80 @@ impl Volume {
...
+        ensures
+            lifecycle_handler@ == LifecycleHandlerView::default(),
+    {
+        LifecycleHandler {
+            inner: deps_hack::k8s_openapi::api::core::v1::LifecycleHandler::default(),
+        }
+    }
+
+    #[verifier(external_body)]
+    pub fn set_exec(&mut self, commands: Vec<String>)
```
marshtompsxd, line 562: Better to name the argument consistent with the field, that is `exec`

src/kubernetes_api_objects/pod.rs
```diff
@@ -1081,6 +1172,46 @@ impl Marshalable for ContainerView {
...
+}
+
+impl LifecycleHandlerView {
+    pub open spec fn default() -> LifecycleHandlerView {
+        LifecycleHandlerView {
+            execs: Option::None,
+        }
+    }
+
+    pub open spec fn set_exec(self, commands: Seq<StringView>) -> LifecycleHandlerView {
```
marshtompsxd, line 1207: Better to name the argument consistent with the field, that is `exec`

src/kubernetes_api_objects/persistent_volume_claim.rs
```diff
@@ -313,4 +313,98 @@ impl Marshalable for PersistentVolumeClaimSpecView {
     proof fn marshal_preserves_integrity() {}
 }
 
+#[verifier(external_body)]
```
marshtompsxd, line 316: I am not sure whether you want to place `StatefulSetPersistentVolumeClaimRetentionPolicy` in this file. Probably better to move it to stateful_set.rs (since it is only used by statefulset)

src/controller_examples/zookeeper_controller/exec/reconciler.rs
```diff
@@ -24,12 +24,17 @@ pub struct ZookeeperReconcileState {
     // reconcile_step, like a program counter, is used to track the progress of reconcile_core
     // since reconcile_core is frequently "trapped" into the controller_runtime spec.
     pub reconcile_step: ZookeeperReconcileStep,
+    pub old_sts: Option<StatefulSet>,
```
marshtompsxd, line 27: The name `old_sts` is a bit weird because there is no new one here. Maybe name it as `sts_from_get`

src/controller_examples/zookeeper_controller/exec/zookeeper_lib/helper_funcs.rs
```diff
@@ -37,6 +37,19 @@ pub fn reconcile_zk_node(path: String, uri: String, replicas: String) -> ZKNodeR
     let path_look_up = path_look_up.unwrap();
     match path_look_up {
         Some(_) => {
+            // Update the cluster size
```
marshtompsxd, line 40: @ZichengMa Could you add more comments in this branch and the next one (that creates the zk node) to educate the future us that why we need to create and update the zk node? Try to write them as technical documentation, rather than person comments

marshtompsxd: Or write the comment in the exec reconcile_core, which is better I believe

marshtompsxd:
@ZichengMa Just to double check: you have tested the zk controller locally, right? I think we should later add the test for scaling down/up into the e2e test and even include them into our CI

ZichengMa:
> @ZichengMa Just to double check: you have tested the zk controller locally, right? I think we should later add the test for scaling down/up into the e2e test and even include them into our CI

Yes, I tested it, but the e2e test has not been finished yet.

src/kubernetes_api_objects/pod.rs
ZichengMa: Currently we only support to use commands in exec. I think we can leave it as 'commands' here and if we extend the exec later, we can modify the name.

Dockerfile
marshtompsxd, line 1: Why do we add the Dockerfile? I guess you copied it to the root directory for convenient use and then mistakenly added it?

kind-with-registry.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/sh
```
marshtompsxd, line 1: What is this script for? Did you write it from scratch?

reference-controllers/deploy.yaml
marshtompsxd, line 1: What is this file for? I think we already have the deploy yaml file under the deploy folder?

reference-controllers/env.sh
marshtompsxd, line 1: Same question as above

reference-controllers/rabbitmq.yaml
marshtompsxd, line 1: Same question as above

reference-controllers/rmqcrd.yaml
marshtompsxd, line 1: Same question, and for all the files below until zookeeperTeardown.sh

src/change_pass.yaml
marshtompsxd, line 1: And what is this one for?

src/cr.yaml
marshtompsxd, line 1: I guess we don't need this as well?

src/current.yaml
marshtompsxd, line 1: Same as above

src/rabbitmq_crd.yaml
marshtompsxd, line 1: I guess we don't need this too?

src/rbmq.yaml
marshtompsxd, line 1: Same as above

src/secret.yaml
marshtompsxd, line 1: Same as above

Dockerfile
ZichengMa: Yes. My bad.

PR #190, opened 2023-08-01
Add simplified specs to the zookeeper external library

euclidgame:
The spec models the data stored in zookeeper as a map whose key is the zk node path. Calling the zk API to write to the zk node will insert the key-value pair into the map. The spec is highly simplified compared to the real zookeeper and is tailored for the use of the zookeeper controller.

There are still many important problems left for future PRs to deal with, such as: (1) what is our assumption on the external APIs? Do we assume they always return the result successfully, or they might return some transient error (and eventually get back to normal)? Or some other assumptions? (2) The external API's state and result might depend on the state of the k8s objects. Like, if the zk pods are not ready then the zk API won't work. Should we try to specify such dependencies?

### DIFF
--- a/src/controller_examples/rabbitmq_controller/proof/safety.rs
+++ b/src/controller_examples/rabbitmq_controller/proof/safety.rs
@@ -211,34 +211,120 @@ pub proof fn lemma_always_at_most_one_create_cm_req_since_rest_id_is_in_flight(
             always(lift_state(at_most_one_create_cm_req_since_rest_id_is_in_flight(key, rest_id)))
         ),
 {
-    lemma_always_filtered_create_cm_req_len_is_at_most_one(spec, key, rest_id);
+    let init = |s: ClusterState| {
+        &&& rest_id_counter_is(rest_id)(s)
+        &&& every_in_flight_msg_has_lower_id_than_allocator()(s)
+        &&& pending_msg_at_after_create_server_config_map_step_is_create_cm_req(key)(s)
+    };
+    let stronger_next = |s, s_prime: ClusterState| {
+        &&& next::<RabbitmqClusterView, RabbitmqReconciler>()(s, s_prime)
+        &&& crash_disabled()(s)
+        &&& busy_disabled()(s)
+        &&& pending_msg_at_after_create_server_config_map_step_is_create_cm_req(key)(s)
+        &&& each_key_in_reconcile_is_consistent_with_its_object()(s)
+        &&& rest_id_counter_is_no_smaller_than(rest_id)(s)
+        &&& every_in_flight_msg_has_unique_id()(s)
+    };
+    let invariant = at_most_one_create_cm_req_since_rest_id_is_in_flight(key, rest_id);
 
-    let p = lift_state(at_most_one_create_cm_req_since_rest_id_is_in_flight(key, rest_id));
-    let q = lift_state(filtered_create_cm_req_len_is_at_most_one(key, rest_id));
-
-    assert_by(
-        p == q, {
-            assert forall |ex| p.satisfied_by(ex) implies q.satisfied_by(ex) by {
-                multiset_lemmas::filtered_size_is_zero_means_no_such_value(
-                    ex.head().network_state.in_flight, cm_create_request_msg_since(key, rest_id)
-                );
-                multiset_lemmas::filtered_size_is_one_means_only_one_such_value(
-                    ex.head().network_state.in_flight, cm_create_request_msg_since(key, rest_id)
-                );
-            }
-
-            assert forall |ex| q.satisfied_by(ex) implies p.satisfied_by(ex) by {
-                multiset_lemmas::filtered_size_is_zero_means_no_such_value(
-                    ex.head().network_state.in_flight, cm_create_request_msg_since(key, rest_id)
-                );
-                multiset_lemmas::filtered_size_is_one_means_only_one_such_value(
-                    ex.head().network_state.in_flight, cm_create_request_msg_since(key, rest_id)
-                );
-            }
-
-            temp_pred_equality(p, q);
-        }
+    entails_and_n!(
+        spec,
+        lift_state(rest_id_counter_is(rest_id)),
+        lift_state(every_in_flight_msg_has_lower_id_than_allocator()),
+        lift_state(pending_msg_at_after_create_server_config_map_step_is_create_cm_req(key))
     );
+    temp_pred_equality(
+        lift_state(init),
+        lift_state(rest_id_counter_is(rest_id))
+        .and(lift_state(every_in_flight_msg_has_lower_id_than_allocator()))
+        .and(lift_state(pending_msg_at_after_create_server_config_map_step_is_create_cm_req(key)))
+    );
+
+    entails_always_and_n!(
+        spec,
+        lift_action(next::<RabbitmqClusterView, RabbitmqReconciler>()),
+        lift_state(crash_disabled()),
+        lift_state(busy_disabled()),
+        lift_state(pending_msg_at_after_create_server_config_map_step_is_create_cm_req(key)),
+        lift_state(each_key_in_reconcile_is_consistent_with_its_object()),
+        lift_state(rest_id_counter_is_no_smaller_than(rest_id)),
+        lift_state(every_in_flight_msg_has_unique_id())
+    );
+    temp_pred_equality(
+        lift_action(stronger_next),
+        lift_action(next::<RabbitmqClusterView, RabbitmqReconciler>())
+        .and(lift_state(crash_disabled()))
+        .and(lift_state(busy_disabled()))
+        .and(lift_state(pending_msg_at_after_create_server_config_map_step_is_create_cm_req(key)))
+        .and(lift_state(each_key_in_reconcile_is_consistent_with_its_object()))
+        .and(lift_state(rest_id_counter_is_no_smaller_than(rest_id)))
+        .and(lift_state(every_in_flight_msg_has_unique_id()))
+    );
+    assert forall |s, s_prime| invariant(s) && #[trigger] stronger_next(s, s_prime) implies invariant(s_prime) by {
+        let pending_msg = s_prime.pending_req_of(key);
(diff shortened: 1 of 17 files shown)


src/controller_examples/rabbitmq_controller/proof/safety.rs
```diff
@@ -211,34 +211,120 @@ pub proof fn lemma_always_at_most_one_create_cm_req_since_rest_id_is_in_flight(
             always(lift_state(at_most_one_create_cm_req_since_rest_id_is_in_flight(key, rest_id)))
         ),
 {
-    lemma_always_filtered_create_cm_req_len_is_at_most_one(spec, key, rest_id);
-
-    let p = lift_state(at_most_one_create_cm_req_since_rest_id_is_in_flight(key, rest_id));
-    let q = lift_state(filtered_create_cm_req_len_is_at_most_one(key, rest_id));
+    let init = |s: ClusterState| {
```
marshtompsxd, line 214: Actually I was about to say that we should refactor these lemmas in this way. Glad that you are already doing so.

src/controller_examples/rabbitmq_controller/spec/reconciler.rs
```diff
@@ -28,6 +28,8 @@ impl Reconciler<RabbitmqClusterView> for RabbitmqReconciler {
     type T = RabbitmqReconcileState;
     type I = ();
     type O = ();
+    type S = ();
```
marshtompsxd, line 31: We probably should find better names for I, O and S. Any ideas?

src/controller_examples/zookeeper_controller/exec/zookeeper_lib/lib.rs
```diff
@@ -5,7 +5,7 @@ use crate::reconciler::exec::external::*;
 use crate::zookeeper_controller::common::*;
 use crate::zookeeper_controller::exec::zookeeper_lib::helper_funcs::*;
 use crate::zookeeper_controller::exec::zookeepercluster::*;
-use crate::zookeeper_controller::spec::zookeeper_lib::*;
+use crate::zookeeper_controller::spec::zookeeper_lib::{ZKSupportInputView, ZKSupportOutputView, ZKNodeResultView};
```
marshtompsxd, line 8: Can we instead call them ZKAPIInput/Output...

src/reconciler/spec/reconciler.rs
```diff
@@ -45,7 +47,11 @@ pub trait Reconciler<#[verifier(maybe_negative)] K: ResourceView>: Sized {
     // Lib should contain method process, it should implement a trait (which should be the spec version of ExternalLibrary).
     // It must be a generic currently. This will cause another round of super annoying refactoring. So currently we just
     // add another method to this trait.
-    open spec fn external_process(input: Self::I) -> Option<Self::O>;
+    // This method consumes the input (which should be computed by reconcile_core) and the current state of the external
+    // library and produces the response and the next state of the library.
+    // Use optional state here because: (1) it's easy to initialize since we don't have to require a default or init method,
+    // (2) some libraries don't need a state to hold information, thus, optional state makes sense. 
+    open spec fn external_process(input: Self::I, state: Option<Self::S>) -> (Option<Self::O>, Option<Self::S>);
```
marshtompsxd, line 54: I am actually not sure about using option here. If the API is stateless, then S can just be `()`, and defining a init method for S as part of `reconciler` should be fine. We can live with the option thing if you think using the option is really better.

src/controller_examples/rabbitmq_controller/spec/reconciler.rs
euclidgame: Of course. But this may make some lines look lengthy :-((

src/reconciler/spec/reconciler.rs
euclidgame: Yes, I am not sure about this either. When I was thinking about adding an init method, I just didn't like the Reconciler trait contains 2 out of 6 methods to describe the external library. But using Option also has some downsides.

src/controller_examples/zookeeper_controller/spec/zookeeper_lib.rs
```diff
@@ -20,4 +20,36 @@ pub struct ZKNodeResultView {
...
+    pub data: Map<StringView, StringView>,
+}
+
+pub open spec fn external_process(input: ZKSupportInputView, state: Option<ZooKeeperState>) -> (Option<ZKSupportOutputView>, Option<ZooKeeperState>) {
+    match input {
+        ZKSupportInputView::ReconcileZKNode(path,uri,replicas) => reconcile_zk_node(path, uri, replicas, state),
+    }
+}
+
+pub open spec fn reconcile_zk_node(
```
marshtompsxd, line 33: The current reconcile_zk_node looks fine to me. And I think we should think hard about two problems:
(1) what is our assumption on the external APIs? Do we assume they always return the result successfully, or they might return some transient error (and eventually get back to normal)? Or some other assumptions?
(2) The external API's state and result might depend on the state of the k8s objects. Like, if the zk pods are not ready then the zk API won't work. Should we try to specify such dependencies?

src/controller_examples/rabbitmq_controller/spec/reconciler.rs
marshtompsxd: That's totally fine

src/controller_examples/zookeeper_controller/spec/zookeeper_lib.rs
euclidgame: (1) This has to do with the specific reconciler. For the zookeeper one, whether to return error result or not is fine. 
(2) Maybe this can be considered when we start to strip the library thing off the controller?

euclidgame:
@marshtompsxd Let me know if we need to further specify the zookeeper library. I think this also depends on the property we want to prove.

src/controller_examples/zookeeper_controller/spec/reconciler.rs
```diff
@@ -27,16 +27,17 @@ pub struct ZookeeperReconciler {}
 
 impl Reconciler<ZookeeperClusterView> for ZookeeperReconciler {
     type T = ZookeeperReconcileState;
-    type I = ZKSupportInputView;
-    type O = ZKSupportOutputView;
+    type LibRequest = ZKAPIInputView;
+    type LibResponse = ZKAPIOutputView;
+    type LibState = ZooKeeperState;
```
marshtompsxd, line 32: Could you rename them to `ExternalAPIInput`, `ExternalAPIOutput`, and `ExternalState`?

src/controller_examples/zookeeper_controller/spec/reconciler.rs
```diff
@@ -48,8 +49,12 @@ impl Reconciler<ZookeeperClusterView> for ZookeeperReconciler {
         reconcile_error(state)
     }
 
-    open spec fn external_process(input: ZKSupportInputView) -> Option<ZKSupportOutputView> {
-        Option::None
+    open spec fn external_process(input: ZKAPIInputView, state: ZooKeeperState) -> (Option<ZKAPIOutputView>, ZooKeeperState) {
+        external_process(input, state)
+    }
+
+    open spec fn library_init_state() -> ZooKeeperState {
```
marshtompsxd, line 56: Could you rename this to `init_external_state`?

src/controller_examples/zookeeper_controller/spec/zookeeper_lib.rs
```diff
@@ -2,22 +2,61 @@
...
+
+impl ZooKeeperState {
+    pub open spec fn default() -> ZooKeeperState {
+        ZooKeeperState {
+            data: Map::empty(),
+        }
+    }
+}
+
+pub open spec fn external_process(input: ZKAPIInputView, state: ZooKeeperState) -> (Option<ZKAPIOutputView>, ZooKeeperState) {
```
marshtompsxd, line 35: Could you rename this to `external_transition`?

src/kubernetes_cluster/spec/controller/common.rs
```diff
@@ -51,6 +52,7 @@ pub open spec fn init_controller_state<K: ResourceView, R: Reconciler<K>>() -> C
     ControllerState {
         ongoing_reconciles: Map::empty(),
         scheduled_reconciles: Map::empty(),
+        external_lib_state: R::library_init_state(),
```
marshtompsxd, line 55: Could you rename this to `external_state`?

src/kubernetes_cluster/spec/controller/common.rs
```diff
@@ -13,14 +13,15 @@ verus! {
...
+    pub external_lib_state: R::LibState,
 }
 
 pub struct OngoingReconcile<K: ResourceView, R: Reconciler<K>> {
     pub triggering_cr: K,
     // pending_req_msg: the request message pending for the handling for k8s api
     // lib_response: the response returned by the external library if a request has been processed by it
     pub pending_req_msg: Option<Message>,
-    pub lib_response: Option<R::O>,
+    pub lib_response: Option<R::LibResponse>,
```
marshtompsxd, line 24: Could you rename this to `pending_external_api_output`?

marshtompsxd:
> @marshtompsxd Let me know if we need to further specify the zookeeper library. I think this also depends on the property we want to prove.

I think the spec is good for now.

src/controller_examples/zookeeper_controller/exec/zookeeper_lib/lib.rs
```diff
@@ -66,19 +68,19 @@ impl ZKSupportOutput {
...
+            ZKAPIOutput::ReconcileZKNode(result) => result,
         }
     }
 }
 
-pub struct ZKSupport {}
+pub struct ZKAPI {}
 
-impl ExternalLibrary<ZKSupportInput, ZKSupportOutput> for ZKSupport {
+impl ExternalLibrary<ZKAPIInput, ZKAPIOutput> for ZKAPI {
```
marshtompsxd, line 78: Could you rename this to `ExternalAPI`?

src/shim_layer/mod.rs
```diff
@@ -42,7 +42,7 @@ where
     K::DynamicType: Default + Eq + Hash + Clone + Debug + Unpin,
     ResourceWrapperType: ResourceWrapper<K> + Send,
     ReconcilerType: Reconciler<ResourceWrapperType, ReconcileStateType, I, O, Lib> + Send + Sync + Default,
-    ReconcileStateType: Send, I: Send + ToView, O: Send + ToView, Lib: ExternalLibrary<I, O>,
+    ReconcileStateType: Send, I: Send + ToView, O: Send + ToView, Lib: ExternalAPI<I, O>,
```
marshtompsxd, line 45: Could you rename the I, O and Lib here? These should be the last.

PR #191, opened 2023-08-01
Port the fluent controller

marshtompsxd:
The fluent controller, built on top of Anvil, behaves almost the same as the reference unverified fluent controller, except that: (1) The FluentBit configuration is passed to the controller through the CR's fields, and (2) The ported controller uses Role and RoleBinding, instead of Cluster and ClusterRoleBinding, because Anvil doesn't support cluster-scoped resources for now.

### DIFF
--- a/deploy/fluent/crd.yaml
+++ b/deploy/fluent/crd.yaml
@@ -8,7 +8,8 @@ spec:
     categories: []
     kind: FluentBit
     plural: fluentbits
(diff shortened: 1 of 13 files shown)


euclidgame:
@marshtompsxd , I have read through the changes. But can we handle this after #192 ?

marshtompsxd:
Yes

PR #192, opened 2023-08-02
Move the external api out of the controller state machine

euclidgame:
Add a field `external_api_state` to the state of the whole system for describing: (1) what inputs and outputs related to the external api are not handled (the `in_flight` field); (2) the state of the external api (`external_api_state` field). It is of type `ExternalAPIState` which basically wraps a set and the user-defined external api state.

`ExternalComm` is used to represent communication between the external api and controller. It has two variants: `Input` is computed by controller and sent to and consumed by external api; `Output` is generated by external api after handling the input. Each invariant also carries a nat which represents the id when the input to the external api is generated.

I don't make it a state machine because it should contain more than one actions and which action to take can't be determined independently (simply by the input). So it is neither like the network state machine which only contains one action nor the other state machines which can determine the action by the input. But I am considering make the external api a state machine, whose state is `ExternalAPIType::State` which is provided by the developer/user, not `ExternalAPIState` which is provided by me.


src/controller_examples/rabbitmq_controller/exec/reconciler.rs
```diff
@@ -38,12 +39,12 @@ impl RabbitmqReconcileState {
 pub struct RabbitmqReconciler {}
 
 #[verifier(external)]
-impl Reconciler<RabbitmqCluster, RabbitmqReconcileState, EmptyMsg, EmptyMsg, EmptyLib> for RabbitmqReconciler {
+impl Reconciler<RabbitmqCluster, RabbitmqReconcileState, EmptyType, EmptyType, EmptyLib> for RabbitmqReconciler {
```
marshtompsxd, line 42: Nice. I was about to ask you to rename these two. Actually, can you also rename `EmptyLib`?

euclidgame: How about `EmptyAPI`?

marshtompsxd: Yes

src/kubernetes_cluster/spec/external_api.rs
```diff
@@ -0,0 +1,62 @@
...
+use crate::state_machine::state_machine::*;
+use crate::temporal_logic::defs::*;
+use vstd::{multiset::*, prelude::*};
+
+verus! {
+
+// This enum is a wrapper for attaching rest id to the input and output of the external api.
+// By doing this, we are able to check whether the output matches the input when conducting continue_reconcile.
+#[is_variant]
+pub enum ExternalComm<Input, Output> {
```
marshtompsxd, line 16: This is a crucial data type so we should document it carefully. Could you revise the comment in this way:
ExternalComm is used to represent xxx. It has two variants: Input is xxx and Output is xxx. Each variant also carries a nat which represents xxx.

btw we should not call the nat here as "rest id" -- ExternalComm is not necessarily REST API call. We can just call these nat as "call_id".

marshtompsxd: And, could you also document this data type briefly in the description?

euclidgame: >  ExternalComm is not necessarily REST API call. We can just call these nat as "call_id".

I can call it call id in the document. But it is also generated by rest_id of the state, right? Or do you think we should create another id?

src/kubernetes_cluster/spec/external_api.rs
```diff
@@ -0,0 +1,62 @@
...
+            } else {
+                s.in_flight.insert(output.get_Some_0())
+            }
+        } else {
+            if input.is_Some() {
+                s.in_flight.remove(input.get_Some_0())
+            } else {
+                s.in_flight
+            }
+        },
```
marshtompsxd, line 57: You can probably refactor it as
```
let after_inserting_output =  if output.is_Some() { s.in_flight.remove(input.get_Some_0()) } else { s.in_flight };
let in_flight_prime =  if input.is_Some() { after_inserting_output.remove(input.get_Some_0()) } else { after_inserting_output };
```

euclidgame: That's what I did before.

src/kubernetes_cluster/spec/controller/controller_runtime.rs
```diff
@@ -51,33 +57,42 @@ pub open spec fn continue_reconcile<K: ResourceView, R: Reconciler<K>>() -> Cont
                 &&& !R::reconcile_error(s.ongoing_reconciles[cr_key].local_state)
                 // Split cases:
                 // (1) there is a pending request which is destined for kubernetes api;
-                // (3) there is no pending request which is destined for kubernetes api and there is a external response;
-                // (4) there is no pending request or external response;
+                // (2) there is a pending external api request (so that we should send a message to the external api) and an external api output
```
marshtompsxd, line 60: Do you mean "a pending external api input"?
If so you can change it to "so that we should feed the input to the external api"

src/kubernetes_cluster/spec/controller/controller_runtime.rs
```diff
@@ -89,71 +104,92 @@ pub open spec fn continue_reconcile<K: ResourceView, R: Reconciler<K>>() -> Cont
                             = (input.rest_id_allocator.allocate().0, Option::Some(controller_req_msg(req, input.rest_id_allocator.allocate().1)));
                         let reconcile_state_prime = OngoingReconcile {
                             pending_req_msg: pending_req_msg,
-                            pending_external_api_output: Option::None,
+                            pending_external_api_input: Option::None,
```
marshtompsxd, line 107: I enjoy this change a lot :-)

src/kubernetes_cluster/spec/external_api.rs
```diff
@@ -0,0 +1,62 @@
...
+    // We use a set to accommodate the input and output of the external api which haven't been processed.
+    pub in_flight: Set<ExternalComm<T::Input, T::Output>>,
+}
+
+pub open spec fn external_api_state_init<T: ExternalAPI>(s: ExternalAPIState<T>) -> bool {
+    &&& s.external_api_state == T::init_state()
+    &&& s.in_flight == Set::<ExternalComm<T::Input, T::Output>>::empty()
+}
+
+pub open spec fn external_api_send_output_and_receive_input<T: ExternalAPI>(
```
marshtompsxd, line 41: Why not just call it `external_api_transition`?

src/kubernetes_cluster/spec/cluster.rs
```diff
@@ -150,31 +154,66 @@ pub open spec fn kubernetes_api_next<K: ResourceView, R: Reconciler<K>>() -> Act
...
 }
 
-pub open spec fn controller_next<K: ResourceView, R: Reconciler<K>>() -> Action<State<K, R>, (Option<Message>, Option<ObjectRef>), ()> {
-    let result = |input: (Option<Message>, Option<ObjectRef>), s: State<K, R>| {
-        let host_result = controller::<K, R>().next_result(
-            ControllerActionInput{recv: input.0, scheduled_cr_key: input.1, rest_id_allocator: s.rest_id_allocator},
+pub open spec fn external_api_next<K: ResourceView, E: ExternalAPI, R: Reconciler<K, E>>() -> Action<State<K, E, R>, ExternalComm<E::Input, E::Output>, ()> {
+    Action {
+        precondition: |input: ExternalComm<E::Input, E::Output>, s: State<K, E, R>| {
+            // For the external api action, a valid input must be contained by the in_flight field of the external_api_state.
```
marshtompsxd, line 160: "must be contained by the in_flight field" -> must exist in the in_flight set

src/kubernetes_cluster/spec/external_api.rs
marshtompsxd: oh wait, I guess that would be a bad idea.

src/external_api/exec.rs
```diff
@@ -16,22 +17,20 @@ pub trait ExternalAPI<Input: ToView, Output: ToView> {
 
 // An empty library that implements External Library.
 // This can be used by those controllers that don't rely on a third-party library.
-// Users can define a reconciler as Reconciler<xx, xx, EmptyMsg, EmptyMsg, EmptyLib>.
-pub struct EmptyLib {}
+// Users can define a reconciler as Reconciler<xx, xx, EmptyType, EmptyType, EmptyAPI>.
```
marshtompsxd, line 20: What are xx? Could you use the actual type name here?

euclidgame: I want to say something like a wildcard. I mean if there's not external api required, the last three generics can be like this, and I don't really care about the first two.

src/kubernetes_cluster/spec/external_api.rs
marshtompsxd: I see. It should be generated by rest_id. Maybe I should rename the top level rest_id to sth else

src/kubernetes_cluster/spec/external_api.rs
marshtompsxd: > ng input and then sending output. But this function is actually used at the controller action, which first considers the output then there may be an input to the external api from the controller output.

makes sense

src/controller_examples/rabbitmq_controller/mod.rs
```diff
@@ -2,5 +2,5 @@
 // SPDX-License-Identifier: MIT
 pub mod common;
 pub mod exec;
-pub mod proof;
+// pub mod proof;
```
marshtompsxd, line 5: Do you plan to leave it for the next PR?

src/controller_examples/zookeeper_controller/mod.rs
```diff
@@ -2,5 +2,5 @@
 // SPDX-License-Identifier: MIT
 pub mod common;
 pub mod exec;
-pub mod proof;
+// pub mod proof;
```
marshtompsxd, line 5: And this one -- I guess fixing this one is harder

src/controller_examples/zookeeper_controller/spec/zookeeper_lib.rs
marshtompsxd, line 1: Actually, could you rename this file to `zookeeper_api.rs`?

src/kubernetes_cluster/mod.rs
```diff
@@ -1,4 +1,4 @@
 // Copyright 2022 VMware, Inc.
 // SPDX-License-Identifier: MIT
-pub mod proof;
+// pub mod proof;
```
marshtompsxd, line 3: I thought you have fixed the lemmas here?

src/kubernetes_cluster/spec/cluster.rs
```diff
@@ -150,31 +154,66 @@ pub open spec fn kubernetes_api_next<K: ResourceView, R: Reconciler<K>>() -> Act
     }
 }
 
-pub open spec fn controller_next<K: ResourceView, R: Reconciler<K>>() -> Action<State<K, R>, (Option<Message>, Option<ObjectRef>), ()> {
-    let result = |input: (Option<Message>, Option<ObjectRef>), s: State<K, R>| {
-        let host_result = controller::<K, R>().next_result(
-            ControllerActionInput{recv: input.0, scheduled_cr_key: input.1, rest_id_allocator: s.rest_id_allocator},
+pub open spec fn external_api_next<K: ResourceView, E: ExternalAPI, R: Reconciler<K, E>>() -> Action<State<K, E, R>, ExternalComm<E::Input, E::Output>, ()> {
```
marshtompsxd, line 157: Could you document this action briefly? Like what I did for `schedule_controller_reconcile` (no need to be that detailed)

src/kubernetes_cluster/spec/cluster.rs
```diff
@@ -282,12 +321,12 @@ pub open spec fn busy_kubernetes_api_rejects_request<K: ResourceView, R: Reconci
 /// This action disallows the kubernetes api to be busy from this point.
 /// This is used to constraint the crash behavior for liveness proof:
```
marshtompsxd, line 322: I should have caught this earlier: "This is used to constraint the crash behavior for liveness proof" -> "This is used to constraint the transient error of Kubernetes APIs for liveness proof"

src/kubernetes_cluster/spec/cluster.rs
```diff
@@ -282,12 +321,12 @@ pub open spec fn busy_kubernetes_api_rejects_request<K: ResourceView, R: Reconci
...
 /// the controller eventually stops rejecting requests.
```
marshtompsxd, line 323: *the requests from the controller eventually stop being rejected by transient error

src/kubernetes_cluster/spec/external_api.rs
marshtompsxd: So which one is better?

src/kubernetes_cluster/spec/cluster.rs
euclidgame: I don't know how to describe the relations with shim_layer.

src/kubernetes_cluster/spec/external_api.rs
euclidgame: Just found that I confused `input` with `output`. `input` should be inserted and `output` should be removed...

src/kubernetes_cluster/spec/cluster.rs
marshtompsxd: No, nothing to do with the shim layer. Just briefly describe what this action models in real-world execution.

src/kubernetes_cluster/spec/external_api.rs
```diff
@@ -43,22 +43,12 @@ pub open spec fn external_api_state_init<T: ExternalAPI>(s: ExternalAPIState<T>)
 }
 
 pub open spec fn external_api_send_output_and_receive_input<T: ExternalAPI>(
-    input: Option<ExternalComm<T::Input, T::Output>>, output: Option<ExternalComm<T::Input, T::Output>>, s: ExternalAPIState<T>
+    output: Option<ExternalComm<T::Input, T::Output>>, input: Option<ExternalComm<T::Input, T::Output>>, s: ExternalAPIState<T>
 ) -> ExternalAPIState<T> {
+    let in_flight_remove_output = if output.is_Some() { s.in_flight.remove(output.get_Some_0()) } else { s.in_flight };
+    let in_flight_prime = if input.is_Some() { in_flight_remove_output.insert(input.get_Some_0()) } else { in_flight_remove_output };
```
marshtompsxd, line 49: Shouldn't you remove the input and insert the output?

euclidgame: Depends on whose input and output. I mean the input and output of the external api. So in the controller action, the output is sent to the controller and should be removed. Contrarily, the input may be generated and inserted.

PR #195, opened 2023-08-04
Use a struct to carry the whole kubernetes cluster

euclidgame:
Move all the things of the kubernetes cluster, incl. states, specs and proofs, into a big struct `Cluster`, which is defined in `crate::kubernetes_cluster::cluster`. For every custom reconciler, say `R: Reconciler<K, E>`, that whats to use the lemmas or specifications inside `kubernetes_cluster`, please first make an alias type:

```
type MyCluster = Cluster<K, E, R>;
```

Then use `MyCluster::` as prefix to call those methods. `MyCluster` is also the cluster state of the system.


src/kubernetes_cluster/spec/cluster.rs
```diff
@@ -385,63 +395,48 @@ pub open spec fn stutter<K: ResourceView, E: ExternalAPI, R: Reconciler<K, E>>()
...
+pub open spec fn next() -> ActionPred<State<K, E, R>> {
+    |s: State<K, E, R>, s_prime: State<K, E, R>| exists |step: Step<K, E>| Self::next_step(s, s_prime, step)
 }
 
 /// We install the reconciler to the Kubernetes cluster state machine spec
 /// TODO: develop a struct for the compound state machine and make reconciler its member
 /// so that we don't have to pass reconciler to init and next in the proof.
-pub open spec fn sm_spec<K: ResourceView, E: ExternalAPI, R: Reconciler<K, E>>() -> TempPred<State<K, E, R>> {
-    lift_state(init::<K, E, R>()).and(sm_partial_spec::<K, E, R>())
+pub open spec fn sm_spec() -> TempPred<State<K, E, R>> {
```
marshtompsxd, line 423: I think we probably no longer need `sm_spec` and `sm_partial_spec`, but let's keep them for now

src/kubernetes_cluster/mod.rs
```diff
@@ -2,3 +2,11 @@
 // SPDX-License-Identifier: MIT
 pub mod proof;
 pub mod spec;
+
+use std::marker::PhantomData;
+
+pub struct Cluster<K, E, R> {
```
marshtompsxd, line 8: Could you document why we need this empty Cluster struct and also the three phantom markers?

euclidgame: Why the empty struct? We use the struct to hold all the functions. It has not fields so it is empty.

src/kubernetes_cluster/proof/controller_runtime_eventual_safety.rs
```diff
@@ -4,19 +4,13 @@
 use crate::external_api::spec::ExternalAPI;
 use crate::kubernetes_api_objects::{common::*, resource::*};
 use crate::kubernetes_cluster::{
-    proof::{
```
marshtompsxd, line 7: Feels so good to remove these imports

src/kubernetes_cluster/mod.rs
marshtompsxd: Yes, I know that. I am basically asking to document:
(1) why we want to move so many spec/proof fn into the Cluster struct
(2) what is the purpose of the three phantom fields (which I believe are used to suppress some Verus warnings)

src/kubernetes_cluster/proof/controller_runtime_safety.rs
```diff
@@ -129,49 +128,49 @@ pub open spec fn every_in_flight_msg_has_unique_id<K: ResourceView, E: ExternalA
     }
 }
 
-pub proof fn lemma_always_every_in_flight_msg_has_unique_id<K: ResourceView, E: ExternalAPI, R: Reconciler<K, E>>()
+pub proof fn lemma_always_every_in_flight_msg_has_unique_id()
```
marshtompsxd, line 131: At some point, we might want to revise these lemmas to pass a `spec` to them and constrain the spec in the precondition.

src/controller_examples/rabbitmq_controller/proof/liveness.rs
```diff
@@ -171,8 +172,8 @@ proof fn invariants_is_stable(rabbitmq: RabbitmqClusterView)
 // Note that different from the above invariants, these do not hold for the entire execution from init.
 // They only hold since some point (e.g., when the rest id counter is the same as rest_id).
 // Some of these invariants are also based on the assumptions.
-spec fn invariants_since_rest_id(rabbitmq: RabbitmqClusterView, rest_id: RestId) -> TempPred<ClusterState> {
-    always(lift_state(RMQCluster::rest_id_counter_is_no_smaller_than(rest_id)))
+spec fn invariants_since_rest_id(rabbitmq: RabbitmqClusterView, rest_id: RestId) -> TempPred<RMQCluster> {
+    always(lift_state(RMQCluster::rest_id_counter_is_no_smaller_than_state_pred(rest_id)))
```
marshtompsxd, line 176: What is this renaming for (_state_pred)?

euclidgame: There’s a method with the same name, originally implemented for State

src/kubernetes_cluster/spec/client/types.rs
marshtompsxd, line 1: We should also rename the one in controller and k8s api to `types.rs` (previously `common.rs`)

src/kubernetes_cluster/spec/network/types.rs
marshtompsxd, line 1: Just curious: why do you decide to separate the types out?

src/kubernetes_cluster/spec/client/types.rs
euclidgame: I have planned to do that. But I thought common is okay because some non-type things might later be added.

src/kubernetes_cluster/spec/network/types.rs
euclidgame: Those types will be imported by Cluster. But to implement the other methods for Cluster you have to import Cluster. This won’t lead to circular import as long as you don’t import the whole module. But I want make the import relations more clear. Makes sense?

src/controller_examples/rabbitmq_controller/proof/liveness.rs
marshtompsxd: Why not we kill the definition of `rest_id_counter_is_no_smaller_than` and for each usage we either inline it or replace it with `rest_id_counter_is_no_smaller_than_state_pred`?

marshtompsxd: And then remove `_state_pred` from its name

euclidgame: That’s just an alternative. But we have to think about how to name this because it’s common to have two methods with the similar functions. That’s also why I keep both of the two without making one of them inline.

marshtompsxd: Since we use the StatePred version a lot and they usually appear at the surface of each lemma (pre and post conditions), we should always assign the better name (the shorter one) to the StatePred version.
For their bool siblings, maybe call it `XXX_bool`?

src/kubernetes_cluster/mod.rs
```diff
@@ -3,20 +3,20 @@
...
-/// specify the generics whenever calling those spec or proof functions.
-/// 
-/// Three markers (PhantomData) are used here to make the three generics not "unused". Unused generics will trigger
-/// compilation errors.
-pub struct Cluster<K, E, R> {
-    _marker1: PhantomData<K>,
-    _marker2: PhantomData<E>,
-    _marker3: PhantomData<R>,
-}
+// The Cluster struct is a container for all the specifications and lemmas regarding the Kubernetes cluster, including
```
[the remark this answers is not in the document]
euclidgame: What's this? Do you look at the latest version?

PR #197, opened 2023-08-04
Support validation rule in the current state machine

euclidgame:
To support the validation rule mechanism, I extend the trait `ResourceView` by adding two new methods: `rule()` and `transition_rule()`. One is for rules that only check the new CR object; the other is for rules that compare the new CR object with the old one. The naming conforms to the [Kubernetes documents](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules). For those types that don't need a validation check, the two methods simply return `true`. I check the validity of the custom resource object in methods `validate_update_request` and `validate_create_request`.

### DIFF
--- a/src/controller_examples/simple_controller/exec/reconciler.rs
+++ b/src/controller_examples/simple_controller/exec/reconciler.rs
@@ -3,7 +3,7 @@
 #![allow(unused_imports)]
 use crate::external_api::exec::*;
 use crate::kubernetes_api_objects::{
-    api_method::*, common::*, config_map::*, object_meta::*, resource::ResourceView,
+    api_method::*, common::*, config_map::*, object_meta::*, resource::*,
 };
 use crate::pervasive_ext::string_map::StringMap;
 use crate::reconciler::exec::{io::*, reconciler::*};
@@ -33,12 +33,12 @@ impl SimpleReconcileState {
 pub struct SimpleReconciler {}
 
 #[verifier(external)]
-impl Reconciler<CustomResource, SimpleReconcileState, EmptyType, EmptyType, EmptyAPI> for SimpleReconciler {
+impl Reconciler<SimpleCR, SimpleReconcileState, EmptyType, EmptyType, EmptyAPI> for SimpleReconciler {
     fn reconcile_init_state(&self) -> SimpleReconcileState {
         reconcile_init_state()
     }
 
-    fn reconcile_core(&self, cr: &CustomResource, resp_o: Option<Response<EmptyType>>, state: SimpleReconcileState) -> (SimpleReconcileState, Option<Request<EmptyType>>) {
+    fn reconcile_core(&self, cr: &SimpleCR, resp_o: Option<Response<EmptyType>>, state: SimpleReconcileState) -> (SimpleReconcileState, Option<Request<EmptyType>>) {
         reconcile_core(cr, resp_o, state)
     }
 
@@ -94,7 +94,7 @@ pub fn reconcile_error(state: &SimpleReconcileState) -> (res: bool)
 // TODO: need to prove whether the object is valid; See an example:
 // ConfigMap "foo_cm" is invalid: metadata.name: Invalid value: "foo_cm": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.',
 // and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
-pub fn reconcile_core(cr: &CustomResource, resp_o: Option<Response<EmptyType>>, state: SimpleReconcileState) -> (res: (SimpleReconcileState, Option<Request<EmptyType>>))
+pub fn reconcile_core(cr: &SimpleCR, resp_o: Option<Response<EmptyType>>, state: SimpleReconcileState) -> (res: (SimpleReconcileState, Option<Request<EmptyType>>))
     requires
         cr@.metadata.name.is_Some(),
         cr@.metadata.namespace.is_Some(),
@@ -126,7 +126,7 @@ pub fn reconcile_core(cr: &CustomResource, resp_o: Option<Response<EmptyType>>, 
     }
 }
 
-pub fn make_configmap(cr: &CustomResource) -> (cm: ConfigMap)
+pub fn make_configmap(cr: &SimpleCR) -> (cm: ConfigMap)
     requires
         cr@.metadata.name.is_Some(),
         cr@.metadata.namespace.is_Some(),
--- a/src/kubernetes_api_objects/cluster_role.rs
+++ b/src/kubernetes_api_objects/cluster_role.rs
@@ -135,17 +135,6 @@ impl ClusterRoleView {
             ..self
         }
     }
-
-    pub closed spec fn marshal_spec(s: ClusterRoleSpecView) -> Value;
-
-    pub closed spec fn unmarshal_spec(v: Value) -> Result<ClusterRoleSpecView, ParseDynamicObjectError>;
-
-    #[verifier(external_body)]
-    pub proof fn spec_integrity_is_preserved_by_marshal()
-        ensures
-            forall |s: ClusterRoleSpecView|
-                Self::unmarshal_spec(#[trigger] Self::marshal_spec(s)).is_Ok()
-                && s == Self::unmarshal_spec(Self::marshal_spec(s)).get_Ok_0() {}
 }
 
 impl ResourceView for ClusterRoleView {
@@ -202,11 +191,18 @@ impl ResourceView for ClusterRoleView {
 
     proof fn from_dynamic_preserves_kind() {}
 
-    open spec fn rule(obj: DynamicObjectView) -> bool {
+    closed spec fn marshal_spec(s: ClusterRoleSpecView) -> Value;
+
+    closed spec fn unmarshal_spec(v: Value) -> Result<ClusterRoleSpecView, ParseDynamicObjectError>;
+
+    #[verifier(external_body)]
+    proof fn spec_integrity_is_preserved_by_marshal() {}
+
+    open spec fn rule(spec: ClusterRoleSpecView) -> bool {
         true
     }
 
-    open spec fn transition_rule(new_cr: DynamicObjectView, old_cr: DynamicObjectView) -> bool {
+    open spec fn transition_rule(new_spec: ClusterRoleSpecView, old_spec: ClusterRoleSpecView) -> bool {
         true
     }
 }
--- a/src/kubernetes_api_objects/cluster_role.rs
+++ b/src/kubernetes_api_objects/cluster_role.rs
@@ -198,11 +198,11 @@ impl ResourceView for ClusterRoleView {
     #[verifier(external_body)]
     proof fn spec_integrity_is_preserved_by_marshal() {}
 
-    open spec fn rule(spec: ClusterRoleSpecView) -> bool {
+    open spec fn rule(obj: ClusterRoleView) -> bool {
         true
     }
 
-    open spec fn transition_rule(new_spec: ClusterRoleSpecView, old_spec: ClusterRoleSpecView) -> bool {
+    open spec fn transition_rule(new_obj: ClusterRoleView, old_obj: ClusterRoleView) -> bool {
         true
     }
 }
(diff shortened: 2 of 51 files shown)


src/controller_examples/simple_controller/exec/reconciler.rs
```diff
@@ -33,12 +33,12 @@ impl SimpleReconcileState {
 pub struct SimpleReconciler {}
 
 #[verifier(external)]
-impl Reconciler<CustomResource, SimpleReconcileState, EmptyType, EmptyType, EmptyAPI> for SimpleReconciler {
```
marshtompsxd, line 36: oh yeah I love this renaming. Planned for long time but never did it

src/kubernetes_api_objects/resource.rs
```diff
@@ -79,4 +79,67 @@ pub trait ResourceView: Sized {
                     ==> d.kind == Self::kind();
 }
 
+pub trait CustomResourceView: Sized {
```
marshtompsxd, line 82: Seems that CustomResourceView repeats a lot of things inside ResourceView. Why not we merge the two rule methods into ResourceView (and not keep CustomResourceView)? For all the built-in types, their rules always return true anyway. This reduces the complexity in terms of traits we need to manage

src/kubernetes_cluster/spec/kubernetes_api/state_machine.rs
```diff
@@ -34,6 +34,8 @@ verus! {
 //
 // + Support more operations like List
 
+impl <K: CustomResourceView, E: ExternalAPI, R: Reconciler<K, E>> Cluster<K, E, R> {
```
marshtompsxd, line 37: I am not against it, but what is the benefit of including all these spec methods into Cluster?

src/kubernetes_cluster/spec/kubernetes_api/state_machine.rs
```diff
@@ -176,6 +180,8 @@ pub open spec fn validate_update_request(req: UpdateRequest, s: KubernetesAPISta
         && req.obj.metadata.resource_version != s.resources[req.key].metadata.resource_version {
         // Update fails because the object has a wrong rv
         Option::Some(APIError::Conflict)
+    } else if req.obj.kind == Kind::CustomResourceKind && !K::transition_rule(req.obj, s.resources.index(req.key)) {
```
marshtompsxd, line 183: You can just do `s.resources[req.key]`

src/kubernetes_cluster/spec/kubernetes_api/state_machine.rs
```diff
@@ -176,6 +180,8 @@ pub open spec fn validate_update_request(req: UpdateRequest, s: KubernetesAPISta
...
+        Option::Some(APIError::IllegalCustomResource)
```
marshtompsxd, line 184: Is it the same error that K8s will return in real execution if validation cannot pass?

marshtompsxd:
@euclidgame have you tried validation rules in a kind k8s cluster? Like, are invalid cr objects indeed rejected by k8s and what is the error like?

src/kubernetes_api_objects/resource.rs
euclidgame: I'd rather wait for verus to support the check for supertrait.

src/kubernetes_cluster/spec/kubernetes_api/state_machine.rs
euclidgame: Otherwise, these spec methods will have to carry generics.

src/kubernetes_api_objects/resource.rs
marshtompsxd: I am not sure whether supertrait is on their roadmap. The problem with this design is that it prevents us from implementing controllers that manage built-in resources, but we might not get there.

euclidgame: You mean in that case the CR type is some built-in type?

marshtompsxd: Another alternative is to wrap the validation related methods into a trait (maybe `Validation`?) and each concrete cr type implement `ResourceView + Validation`

marshtompsxd: > You mean in that case the CR type is some built-in type?

Yes. People usually define their own CR type for controllers but sometimes they just manage the built-in types.

euclidgame:
@marshtompsxd , here it is:
```
❯ kubectl apply -f deploy/rabbitmq/rabbitmq.yaml                                                                                                 
The RabbitmqCluster "rabbitmq" is invalid: spec.replicas: Invalid value: "integer": scaling down is not supported
```

> scaling down is not supported

This is defined by me (or see `deploy.yaml` of rabbitmq please)

src/kubernetes_api_objects/cluster_role.rs
```diff
@@ -201,6 +201,14 @@ impl ResourceView for ClusterRoleView {
     proof fn from_dynamic_preserves_metadata() {}
 
     proof fn from_dynamic_preserves_kind() {}
+
+    open spec fn rule(obj: DynamicObjectView) -> bool {
```
marshtompsxd, line 205: I really hope default trait method can land soon

src/kubernetes_api_objects/config_map.rs
```diff
@@ -231,6 +231,14 @@ impl ResourceView for ConfigMapView {
     proof fn from_dynamic_preserves_metadata() {}
 
     proof fn from_dynamic_preserves_kind() {}
+
+    open spec fn rule(obj: DynamicObjectView) -> bool {
+        true
+    }
+
+    open spec fn transition_rule(new_cr: DynamicObjectView, old_cr: DynamicObjectView) -> bool {
```
marshtompsxd, line 239: I think you can just rename the two arguments to `new_obj` and `old_obj` for all types

src/kubernetes_api_objects/resource.rs
```diff
@@ -77,6 +77,10 @@ pub trait ResourceView: Sized {
             forall |d: DynamicObjectView|
                 #[trigger] Self::from_dynamic_object(d).is_Ok()
                     ==> d.kind == Self::kind();
+    
+    open spec fn rule(obj: DynamicObjectView) -> bool;
```
marshtompsxd, line 81: Actually I think rule and transtion_rule should take Self, instead of DynamicObjectView. Instead, k8s api state machine should do the unmarshalling work (since we will reject all the DynamicObjects that cannot be unmarshalled correctly at k8s api state machine).

src/kubernetes_cluster/spec/kubernetes_api/state_machine.rs
```diff
@@ -87,12 +89,14 @@ pub open spec fn validate_create_request(req: CreateRequest, s: KubernetesAPISta
...
         // Creation fails because the namespace of the provided object does not match the namespace sent on the request
         Option::Some(APIError::BadRequest)
-    } else if !object_has_well_formed_spec(req.obj) {
+    } else if !Self::object_has_well_formed_spec(req.obj) {
         // Creation fails because the spec of the provided object is not well formed
         Option::Some(APIError::BadRequest) // TODO: should the error be BadRequest?
     } else if s.resources.dom().contains(req.obj.set_namespace(req.namespace).object_ref()) {
         // Creation fails because the object already exists
         Option::Some(APIError::ObjectAlreadyExists)
+    } else if req.obj.kind == Kind::CustomResourceKind && !K::rule(req.obj) {
```
marshtompsxd, line 98: I would suggest let's unmarshal the object here (K::from_dynamic) and pass the unmarshalled object to rule (and transition rule).

src/kubernetes_cluster/spec/kubernetes_api/state_machine.rs
```diff
@@ -176,6 +180,8 @@ pub open spec fn validate_update_request(req: UpdateRequest, s: KubernetesAPISta
         && req.obj.metadata.resource_version != s.resources[req.key].metadata.resource_version {
         // Update fails because the object has a wrong rv
         Option::Some(APIError::Conflict)
+    } else if req.obj.kind == Kind::CustomResourceKind && !K::transition_rule(req.obj, s.resources[req.key]) {
```
marshtompsxd, line 183: Don't we need to apply `rule` for update request as well? I think k8s will check `rule` for both create and update.

marshtompsxd:
> we have no idea of what the kind of the object got from ectd via req.key, thus unable to judge whether the result of unmarshal of the oldSelf is ok

I think the problem here is that we don't check whether the dynamic object can be successfully unmarshalled into `K` in the current k8s api state machine. We should add that check, and then pass `K`, instead of DynamicObject, to rules and transition rules.

src/kubernetes_cluster/spec/kubernetes_api/state_machine.rs
```diff
@@ -34,6 +34,8 @@ verus! {
 //
 // + Support more operations like List
 
+impl <K: ResourceView, E: ExternalAPI, R: Reconciler<K, E>> Cluster<K, E, R> {
+
 // TODO: maybe make it a method of DynamicObjectView?
 pub open spec fn object_has_well_formed_spec(obj: DynamicObjectView) -> bool {
```
marshtompsxd, line 40: Can we add `&&& obj.kind == CustomResource::kind() ==> K::unmarshal_spec(obj.spec).is_Ok()` here?

marshtompsxd:
Also note that the design in this PR allows users to specify whatever types of validation rules, which might not be supported by k8s. We might need to find a way to constrain that: ideally, Anvil should prepare the predicates for all the supported validation rules and the user can and have to compose these predicates to get their `rule` and `transition_rule`.

euclidgame:
> We might need to find a way to constrain that: ideally, Anvil should prepare the predicates for all the supported validation rules

@marshtompsxd  First, I don't think it's realistic to do this and in fact, CEL already accounts for a large amount of predicates and is extensible; Second, typically users only write rules for CR and they should make the spec one match the rules they write in the yaml. Thus, you don't have to constraint it in the state machine.

src/kubernetes_cluster/spec/kubernetes_api/state_machine.rs
euclidgame: That's not the problem.

euclidgame: What I need is the kind of object conforms to the one of object in `etcd`. Adding such a constraint only tells that the new object can be successfully `unmarshall`ed.

marshtompsxd: But the old one can also be successfully unmarshalled because the old request was also checked whether it was well formed. And we have the invariant saying that each object in etcd can be unmarshalled (that invariant doesn't include cr objects but it can be included). So in the proof it is easy to show that old object can also be unmarshalled.

euclidgame: What does this have to do with proof? We need to pass the result of unmarshal in the state machine to `transition_rule`.

marshtompsxd: In the spec, nothing prevents you from calling `K::unmarshall(old_obj.spec).get_Ok_0()` and you can pretend that unmarshall always returns an Ok in the spec code.
The only problem is whether we can show that unmarshall indeed returns OK for old_object to the verifier, and that invariant mentioned above solves this problem.

euclidgame: It's not pretend.... I guess what you want to do is just to return an error when the unmarshal process of old_obj fails.

marshtompsxd: Sorry I don't get you. I am just proposing to write
```
transition_rule(K::from_dynamic_object(old_obj).get_Ok_0(), K::from_dynamic_object(new_obj).get_Ok_0())
```
What would be your concerns here?

euclidgame: Sorry, maybe I misunderstand something. But should we move `marshal` and `unmarshal` to `ResourceView`?

marshtompsxd:
OK now I see the point. Sorry that I missed it. Let's leave it as it is in this PR and move it to the trait in the next PR
________________________________

euclidgame:
It doesn't matter. Already finish it.

marshtompsxd:
Great work

marshtompsxd:
Can you also update the description? I think it is a bit outdated

PR #200, opened 2023-08-07
Document garbage collector

marshtompsxd:
This is used as later reference for us to better specify the behavior of the garbage collector.

PR #201, opened 2023-08-07
Extend cluster state machine with the garbage collector specification

marshtompsxd:
Make the cluster state machine specification more extensible by adding a new action that models the behavior of built-in controllers in Kubernetes, and use this new action to specify the behavior of the garbage collector.

In Kubernetes, built-in controllers also follow the state-reconciliation principle (as the custom controllers), but they are implemented in a different style compared to the `kube-rs` style (all logic implemented in a `reconcile` function). The built-in controllers set up different informers to watch different from the API server, form a local view of the cluster state, and then update the cluster state depending on its local view. For example, the garbage collector listens to a broad range of events in the Kubernetes cluster, and build a dependency graph for the ownership relationship between the objects, and decides which object should be garbage collected (because its owners are all gone).

Precisely specifying all the behavior above is tedious and unnecessary: it requires us to specify the communication between the informer and the API server and how each controller, and reason about it during the proof. Instead of specifying all the details, we focus on the core state-reconciliation logic in each built-in controller. That is, how the controller would update the cluster state given the current cluster state.

More concretely, we make each built-in controller action take two inputs from the top-level cluster state machine: the current cluster state (from the API server state machine), and one object reference that points to the object the controller should manage (this is similar to how we trigger the custom controller in `schedule_controller_reconcile`). Besides, since all the built-in controllers are folded into one action of the cluster state machine, the cluster state machine also chooses which built-in controller to run so that our weak fairness assumption on this action ensures each built-in controller eventually makes some progress.

In this way, we specify the behavior of the garbage collector. Note that the current specification is still highly simplified compared to its actual behavior (see the documentation in #200 ): we miss the specification for foreground deletion and orphan dependents, which are left for future work. Currently, the specified garbage collector will just delete an object if all of its owners are not present in the cluster state.

After introducing the garbage collector, some of the lemmas that talk about the absence of deletion to some objects will fail. Fixing them should not be difficult but we will need to update the reconciliation logic to set owner_references to these objects, hereby the lemma will need fix again. We will fix these lemmas after we finish updating the reconciliation logic.

Finally, this PR also removes some methods in OwnerReference and its view for setting different fields. This is because of the embarrassing difference between the data type of the actual `kind` and `uid` (which should be String) and the data type used in our specification (enum and nat) -- we intentionally decide to use different data types for the ease of proof and specification. As another part of future work, we will instead implement a trusted method for the CR objects that returns an owner reference that points to the CR object and uses it in the new reconciliation implementation. This trusted method will have the postcondition that the returned view of the owner reference shares the same `kind` and `uid` with its own view.

### DIFF
--- a/src/kubernetes_cluster/spec/builtin_controllers/garbage_collector.rs
+++ b/src/kubernetes_cluster/spec/builtin_controllers/garbage_collector.rs
@@ -46,7 +46,7 @@ pub open spec fn run_garbage_collector() -> BuiltinControllersAction {
                 ||| !resources.dom().contains(owner_reference_to_object_reference(owner_references[i], namespace))
                 // or it exists but has a different uid
                 // (which means the actual owner was deleted and another object with the same name gets created again)
-                ||| resources[owner_reference_to_object_reference(owner_references[i], namespace)].metadata.uid.get_Some_0() != owner_references[i].uid
+                ||| resources[owner_reference_to_object_reference(owner_references[i], namespace)].metadata.uid != Some(owner_references[i].uid)
             } {
                 let delete_req_msg = built_in_controller_req_msg(delete_req_msg_content(
                     key, input.rest_id_allocator.allocate().1


euclidgame:
Can we first set the owner_references and then fix the lemmas?

marshtompsxd:
Yes that's the plan

src/kubernetes_cluster/spec/builtin_controllers/garbage_collector.rs
```diff
@@ -0,0 +1,64 @@
...
+            let namespace = resources[key].metadata.namespace.get_Some_0();
+            let owner_references = resources[key].metadata.owner_references.get_Some_0();
+            // The garbage collector decides whether to delete an object by checking its owner references,
+            // it deletes the object if for each owner reference...
+            if forall |i| #![trigger owner_references[i]] {
+                // the referred owner object does not exist in the cluster state
+                ||| !resources.dom().contains(owner_reference_to_object_reference(owner_references[i], namespace))
+                // or it exists but has a different uid
+                // (which means the actual owner was deleted and another object with the same name gets created again)
+                ||| resources[owner_reference_to_object_reference(owner_references[i], namespace)].metadata.uid.get_Some_0() != owner_references[i].uid
```
euclidgame, line 49: Do you think it's better to say
```
||| resources[owner_reference_to_object_reference(owner_references[i], namespace)].metadata.uid != Some(owner_references[i].uid)
```

marshtompsxd: updated

euclidgame (approved):

PR #204, opened 2023-08-09
Add github workflow for e2e test

ZichengMa:
Add a new yaml file in workflow called e2e-testing.yaml. It will build and deploy two controllers and then run e2e test for each of them.


.github/workflows/e2e-testing.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Push rabbitmq controller image to local registry
+        run: docker push localhost:5001/rabbitmq-controller:1.0
+      - name: Deploy zookeeper controller
+        run: ./e2e/deploy.sh zookeeper
+      - name: Deploy rabbitmq controller
+        run: ./e2e/deploy.sh rabbitmq
+      - name: Run zookeeper e2e tests
+        run: cd e2e && cargo run -- zookeeper
+      - name: Run rabbitmq e2e tests
+        run: cd e2e && cargo run -- rabbitmq
```
marshtompsxd, line 48: I would suggest we test the controllers separately: create kind cluster -> test zk -> teardown kind cluster -> create kind cluster -> test rabbitmq -> teardown cluster

e2e/deploy/kind-with-registry.sh
```diff
@@ -0,0 +1,69 @@
...
+metadata:
+  name: local-registry-hosting
+  namespace: kube-public
+data:
+  localRegistryHosting.v1: |
+    host: "localhost:${reg_port}"
+    help: "https://kind.sigs.k8s.io/docs/user/local-registry/"
+EOF
+
+docker pull kindest/node:v1.27.3
```
marshtompsxd, line 69: You are using v1.27.3 which is probably a newer version compared to what you used before -- this might be the reason why you need extra rules for Role/Rolebinding (the version I have been using is v1.26.3).
We can use this version(v1.27.3) for the e2e test at least for now.

marshtompsxd: oh wait, I am actually a bit confused. Why do you pull the image here?

e2e/deploy/kind-with-registry.sh
marshtompsxd, line 1: @ZichengMa Did you refer to some other materials when writing this script? If so, could you send me a link?

marshtompsxd: It seems to me a large part of the script is trying to connect a local container registry to the kind cluster, because you want to build new images. There is a easier way: you can just build the image and than load the image to each node in the kind cluster using https://kind.sigs.k8s.io/docs/user/quick-start/#loading-an-image-into-your-cluster. Besides, you need to update the image pull policy in your deployment yaml file to prefer local image first https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy

e2e/deploy/kind-with-registry.sh
```diff
@@ -0,0 +1,69 @@
...
+
+# 2. Create kind cluster with containerd registry config dir enabled
+# TODO: kind will eventually enable this by default and this patch will
+# be unnecessary.
+#
+# See:
+# https://github.com/kubernetes-sigs/kind/issues/2875
+# https://github.com/containerd/containerd/blob/main/docs/cri/config.md#registry-configuration
+# See: https://github.com/containerd/containerd/blob/main/docs/hosts.md
+cat <<EOF | kind create cluster --config=-
```
marshtompsxd, line 21: When you create the kind cluster, you can specify multiple worker nodes (say 1 control plane node and 3 worker nodes), which can make the test run faster

e2e/deploy/deploy_rabbitmq.yaml
```diff
@@ -0,0 +1,103 @@
...
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: rabbitmq-controller
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: rabbitmq-controller
+    spec:
+      containers:
+        - image: localhost:5001/rabbitmq-controller:1.0
```
marshtompsxd, line 98: Minor: if you want to use a version number as tag, use `v0.1.0`

e2e/deploy/deploy_zookeeper.yaml
```diff
@@ -0,0 +1,94 @@
...
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: zookeeper-controller
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: zookeeper-controller
+    spec:
+      containers:
+        - image: localhost:5001/zookeeper-controller:1.0
```
marshtompsxd, line 89: Minor: if you want to use a version number as tag, use `v0.1.0`

e2e/deploy/kind-with-registry.sh
ZichengMa: https://kind.sigs.k8s.io/docs/user/local-registry/ This is the reference. I used local registry to load our image.

marshtompsxd:
@ZichengMa Can you fix the failure, and just load the images to the kind node, instead of creating a local registry? If possible, we should keep the workflow simple.

e2e/deploy.sh
marshtompsxd, line 1: Is this script still needed?

e2e/deploy/deploy_rabbitmq.yaml
marshtompsxd, line 1: Same as above

e2e/deploy/deploy_zookeeper.yaml
marshtompsxd, line 1: Same as above

e2e/deploy/deploy_rabbitmq.yaml
ZichengMa: Will remove it.

PR #207, opened 2023-08-10
Hold deletion if finalizers are present

marshtompsxd:
Update the API server state machine to check the presence of the finalizer before completing the deletion: if the finalizer exists, it sets the deletion timestamp to the object and doesn't delete the object. The deletion is completed by a later update that removes the finalizers from the object, which is left for future work.

Also delete some lemmas that no longer hold (e.g., a deletion request leads to the object getting deleted) with the update state machine.


src/kubernetes_cluster/spec/kubernetes_api/state_machine.rs
```diff
@@ -199,22 +228,25 @@ pub open spec fn handle_update_request(msg: Message, s: KubernetesAPIState) -> (
...
     } else {
-        // Update succeeds
-        // Updates the resource version of the object
-        let updated_obj = req.obj.set_namespace(req.key.namespace).set_resource_version(s.resource_version_counter);
+        // Update succeeds.
+        // Updates the resource version of the object.
+        let old_obj = s.resources[req.key];
+        let updated_obj = req.obj.set_namespace(req.key.namespace)
+                            .set_resource_version(s.resource_version_counter)
+                            .set_uid(old_obj.metadata.uid.get_Some_0());
```
euclidgame, line 247: Why do we have to set the `uid` if the old and new uid are required to be the same in `validate_update_request`?

marshtompsxd: We only require the uid to be the same when `obj.metadata.uid.is_Some()`. If the uid is not present in the metadata, we will set it to the existing uid. See this: https://github.com/kubernetes/kubernetes/blob/v1.26.3/staging/src/k8s.io/apiserver/pkg/registry/rest/update.go#L133

euclidgame: Didn't see `is_Some()`.

PR #211, opened 2023-08-13
Revise the rabbitmq controller to prevent unexpected downscale

euclidgame:
The current version of RabbitMQ controller doesn't support scaling down; directly reducing the replicas of the stateful set can lead to quorum loss and even data loss. More details in [this issue](https://github.com/rabbitmq/cluster-operator/issues/223). Thus, we decide to disallow downscale for the stateful set in our implementation.

 Using the [validation rule](https://github.com/vmware-research/verifiable-controllers/blob/f5236647bf4fb26daa1359fde3c61a282a886735/src/controller_examples/rabbitmq_controller/spec/rabbitmqcluster.rs#L108) can guarantee that updating the deployment won't decrease the replicas. But here is a corner case: a workaround for downscale operation is to delete the current deployment and create a new one with fewer replicas, which doesn't violate the validation rule. If the execution happens in this way, chances are that the stateful set created from the new cr may not have been deleted by the garbage collector when the controller tries to update the stateful set with the new cr which has a smaller `replicas` field. Thus, the controller implementation still needs to compare the old and new replicas before updating stateful set to make sure scaling down doesn't happen. This makes the proof a lot more difficult, because we have to show that the replicas of old stateful set is lower than that of the current cr, whose proof requires us to show that if a stateful set has owner reference pointing to some cr, its replicas is no larger. 

Therefore, we decide to let the controller wait for the garbage collector to delete the old stateful set, which avoids the corner case, and does not introduce too much complexity to the proof since it's needless to compare the stateful set now. In this case, if the old stateful set doesn't have an owner reference pointing to the current cr, the reconcile will simply go to the error state and wait for the next round of reconcile.

To make the liveness proof work, I change `the_object_in_reconcile_has_spec_as` into `the_object_in_reconcile_has_spec_and_uid_as` so that the owner_references can also be derived from the desired custom resource. The left work is as follows:
1. Add an eventual safety property showing that `spec /\ []desired_state_is(cr) |= true ~> (sts_key_exists => sts.owner_references.contains(cr.owner_ref()))`.
2. Add `[](sts_key_exists => sts.owner_references.contains(cr.owner_ref()))` to the assumption of the spec.
3. Add reasoning about the steps after create/update server config map and the stateful set part should be similar as zookeeper as long as 1 and 2 are done.


.gitignore
```diff
@@ -9,3 +9,4 @@ src/*.long-type-*.txt
 src/.verus-log/
 reference-controllers/*-controller/target/
 e2e/target/
+e2e_test/target/
```
marshtompsxd, line 12: No need to add this line. You can just remove `e2e_test/` on your laptop

marshtompsxd:
@euclidgame Thanks for the efforts. This is a super important PR. I will need to review it super carefully.

Can you revise the description as follows:
Rabbitmq controller does not support downscale, and why --> Start with we already use the validation rule to prevent downscale --> But there is still a corner case that downscale can happen, and explain the corner case --> thus, the controller implementation still needs to compare the old and new replicas before updating --> this makes the proof a lot more difficult, and why --> we decide to let the controller wait for the gc to delete the old statefulset, which avoids the corner case, and does not introduce too much complexity to the proof

tianyin:
"super important PR" :)

src/controller_examples/rabbitmq_controller/exec/reconciler.rs
```diff
@@ -298,21 +298,19 @@ pub fn reconcile_core(rabbitmq: &RabbitmqCluster, resp_o: Option<Response<EmptyT
                     if found_stateful_set.is_ok(){
                         let mut new_stateful_set = found_stateful_set.unwrap();
                         // rabbitmq controller doesn't support scale down, so new replicas must be greater than or equal to old replicas
```
marshtompsxd, line 300: We can remove this comment

src/controller_examples/rabbitmq_controller/exec/reconciler.rs
```diff
@@ -298,21 +298,19 @@ pub fn reconcile_core(rabbitmq: &RabbitmqCluster, resp_o: Option<Response<EmptyT
                     if found_stateful_set.is_ok(){
                         let mut new_stateful_set = found_stateful_set.unwrap();
```
marshtompsxd, line 299: The term `new_stateful_set` is a bit confusing. Can we term it as `found_stateful_set` and the previous `found_stateful_set` becomes `get_sts_result`?

src/controller_examples/rabbitmq_controller/exec/reconciler.rs
```diff
@@ -298,21 +298,19 @@ pub fn reconcile_core(rabbitmq: &RabbitmqCluster, resp_o: Option<Response<EmptyT
...
-                                    name: stateful_set.metadata().name().unwrap(),
-                                    namespace: rabbitmq.namespace().unwrap(),
-                                    obj: new_stateful_set.to_dynamic_object(),
-                                });
-                                let state_prime = RabbitmqReconcileState {
-                                    reconcile_step: RabbitmqReconcileStep::AfterUpdateStatefulSet,
-                                    ..state
-                                };
-                                return (state_prime, Some(Request::KRequest(req_o)));
+                        if new_stateful_set.metadata().owner_references_contains(rabbitmq.controller_owner_ref()) {
```
marshtompsxd, line 301: And could you document why we want to check the owner reference here (the corner case)?

src/controller_examples/rabbitmq_controller/exec/reconciler.rs
euclidgame: I also don't know why it was named this way.

src/controller_examples/rabbitmq_controller/proof/common.rs
```diff
@@ -91,28 +114,28 @@ pub open spec fn exists_resp_in_flight_at_rabbitmq_step_with_rabbitmq(
...
     &&& msg.src == HostId::CustomController
     &&& msg.dst == HostId::KubernetesAPI
     &&& msg.content.is_APIRequest()
-    &&& is_correct_pending_request_at_rabbitmq_step(step, msg.content.get_APIRequest_0(), rabbitmq, object)
+    &&& is_correct_pending_request_at_rabbitmq_step(step, msg.content.get_APIRequest_0(), rabbitmq)
 }
 
 pub open spec fn is_correct_pending_request_at_rabbitmq_step(
-    step: RabbitmqReconcileStep, request: APIRequest, rabbitmq: RabbitmqClusterView, object: DynamicObjectView
+    step: RabbitmqReconcileStep, request: APIRequest, rabbitmq: RabbitmqClusterView
```
marshtompsxd, line 138: Nice!

src/controller_examples/rabbitmq_controller/proof/helper_invariants.rs
```diff
@@ -0,0 +1,627 @@
...
+    }
+}
+
+// TODO: fix this lemma.
+// If the configmap has no owner_reference, fixing the lemma is simple:
+// we just need to show that the configmap in the cluster state never has any owner reference.
+//
+// However, later we are going to set the owner_reference of the configmap to the CR object,
+// so we will need the assumption that "CR always exists" to prove this invariant.
+#[verifier(external_body)]
```
marshtompsxd, line 584: This is left for the next PR?

src/controller_examples/rabbitmq_controller/proof/liveness.rs
```diff
@@ -744,17 +763,17 @@ proof fn lemma_from_pending_req_in_flight_at_some_step_to_pending_req_in_flight_
         // next_step != RabbitmqReconcileStep::Init,
         forall |rabbitmq_1, resp_o|
```
marshtompsxd, line 764: Could you also document on why we need this forall? You explained to me several times but sometimes I still forget and hope there is some doc here. Also, you can rename `rabbitmq_1` to `any_rabbitmq`

src/controller_examples/rabbitmq_controller/proof/liveness.rs
```diff
@@ -744,17 +763,17 @@ proof fn lemma_from_pending_req_in_flight_at_some_step_to_pending_req_in_flight_
         // next_step != RabbitmqReconcileStep::Init,
```
marshtompsxd, line 763: You can remove it if not needed

src/controller_examples/rabbitmq_controller/proof/liveness.rs
```diff
@@ -912,21 +929,21 @@ proof fn lemma_from_resp_in_flight_at_some_step_to_pending_req_in_flight_at_next
         // We only know that rabbitmq_1.object_ref() == rabbitmq.object_ref() && rabbitmq_1.spec == rabbitmq.spec.
         forall |rabbitmq_1, resp_o|
```
marshtompsxd, line 930: similar comments as above

src/controller_examples/rabbitmq_controller/spec/rabbitmqcluster.rs
```diff
@@ -94,7 +105,8 @@ impl ResourceView for RabbitmqClusterView {
     }
 
     open spec fn transition_rule(new_obj: RabbitmqClusterView, old_obj: RabbitmqClusterView) -> bool {
-        true
+        new_obj.spec.replicas >= old_obj.spec.replicas
+        // true
```
marshtompsxd, line 109: You can remove the comment

src/kubernetes_cluster/proof/controller_runtime_safety.rs
```diff
@@ -18,235 +18,77 @@ verus! {
...
-            ==> msg.content.get_rest_id() < s.rest_id_allocator.rest_id_counter
+        forall |key: ObjectRef| 
+        #[trigger] s.controller_state.scheduled_reconciles.contains_key(key)
+        && key.kind.is_CustomResourceKind()
+        ==> s.controller_state.scheduled_reconciles[key].metadata().uid.is_Some()
+        && s.controller_state.scheduled_reconciles[key].metadata().uid.get_Some_0() < s.kubernetes_api_state.uid_counter
     }
 }
 
-pub proof fn lemma_always_every_in_flight_msg_has_lower_id_than_allocator
```
marshtompsxd, line 29: What are these changes for? We no longer need these lemmas?

src/controller_examples/rabbitmq_controller/proof/helper_invariants.rs
euclidgame: My plan is that if you don't review this PR immediately, I will try to fix those lemmas. Now that you are reviewing it, we can leave it for next PR.

src/kubernetes_cluster/proof/controller_runtime_safety.rs
marshtompsxd: oh I see. You moved them

src/kubernetes_cluster/proof/cluster.rs
```diff
@@ -125,6 +125,7 @@ pub open spec fn desired_state_is(cr: K) -> StatePred<Self>
         &&& s.resource_key_exists(cr.object_ref())
         &&& K::from_dynamic_object(s.resource_obj_of(cr.object_ref())).is_Ok()
         &&& K::from_dynamic_object(s.resource_obj_of(cr.object_ref())).get_Ok_0().spec() == cr.spec()
+        &&& K::from_dynamic_object(s.resource_obj_of(cr.object_ref())).get_Ok_0().metadata().uid == cr.metadata().uid
```
marshtompsxd, line 128: Actually, we don't have to change the desired_state_is predicate -- the fact that the cr always exists implies that its uid never changes. Could you leave a TODO here that later we will revert this predicate back?

src/controller_examples/rabbitmq_controller/proof/helper_invariants.rs
marshtompsxd: OK

marshtompsxd:
I was looking for the reasoning for the statefulset and then I just realize that the proof is not done yet. Could you also summarize a list of the cosmetic changes made to the proof in the description (revise the last paragraph)?

src/kubernetes_cluster/proof/cluster.rs
euclidgame: I don't understand. Do you mean `[]cr_key_exists |= K::from_dynamic_object(s.resource_obj_of(cr.object_ref())).get_Ok_0().metadata().uid == cr.metadata().uid`?

The `cr` here can be an arbitrary object, `[]cr_key_exists` only shows that the uid of the cr object stored in etcd never changes, not the parameter `cr` here.

marshtompsxd: No, we don't want to say the uid is the same as `cr.metadata().uid`.
I was saying that we should treat uid similarly to the rest_id in the proof.
Basically, we no longer assume `K::from_dynamic_object(s.resource_obj_of(cr.object_ref())).get_Ok_0().metadata().uid == cr.metadata().uid`. Instead, you carry an `uid` as an argument (like the `rest_id`) in whatever lemma that needs the uid, with an extra precondition saying that `[]K::from_dynamic_object(s.resource_obj_of(cr.object_ref())).get_Ok_0().metadata().uid == uid`.
Then, you can prove that whatever `uid`, if initially we have `K::from_dynamic_object(s.resource_obj_of(cr.object_ref())).get_Ok_0().metadata().uid == uid`, then it is always true that `K::from_dynamic_object(s.resource_obj_of(cr.object_ref())).get_Ok_0().metadata().uid == uid` (that is, uid remains unchanaged).
Finally, you can use `leads_to_exists_intro` to eliminate the `uid`. Just like https://github.com/vmware-research/verifiable-controllers/blob/main/src/controller_examples/zookeeper_controller/proof/liveness/liveness_property.rs#L208-L238

marshtompsxd: btw, I just realized that we should later revise `pub open spec fn desired_state_is(cr: K) -> StatePred<Self>` to `pub open spec fn desired_state_is(key: ObjectRef, spec: K::Spec) -> StatePred<Self>`. We only need the key and the spec here.

marshtompsxd: btw2, can we revise `K::from_dynamic_object(s.resource_obj_of(cr.object_ref())).get_Ok_0().metadata().uid` to `s.resource_obj_of(cr.object_ref()).metadata.uid`?

marshtompsxd: > btw, I just realized that we should later revise `pub open spec fn desired_state_is(cr: K) -> StatePred<Self>` to `pub open spec fn desired_state_is(key: ObjectRef, spec: K::Spec) -> StatePred<Self>`. We only need the key and the spec here. I didn't do that because associated type was not there.

Well, this could involve A LOT OF changes... Let's table it and leave a todo there maybe.

euclidgame: > No, we don't want to say the uid is the same as cr.metadata().uid

This doesn't make sense to me. If the uid is not same as cr.metadata().uid, why should we care about that cr? I just think that we should only prove the liveness property about cr that always exists, so I think `desired_state_is` should say that that cr exists in ectd except the status thing you mentioned could be different.

marshtompsxd: > why should we care about that cr?

I mean, we don't need to say that in the `desired_state_is()` (we want to minimize our assumption, if part of the predicate can be derived by the other parts), but I think it does no harm to keep it for now

marshtompsxd: nvm, let's also table this for now

euclidgame: Minimizing the assumption usually means getting a stronger property. But if strengthening the property is not useful, why should we minimize that?

marshtompsxd: I am confused: if the assumption can be simpler (weaker) then why is it a bad thing? Or are you worried that weakening the assumption (for just one line) means many changes in the proof, so it is not worthwhile -- that I understand and agree.

PR #212, opened 2023-08-13
Cosmetic changes in API server spec

marshtompsxd:
This PR makes some small changes to further complete the API server specification:
- Specify unconditional update which decides whether the object can be updated without providing a resource version.
- Revise the conditions on checking whether an update is a noop-update (which does not change the resource version).

There could be more changes to the API server specifications later when we support more fields.


euclidgame:
Where is the source of this change in k8s api?

marshtompsxd:
It is a detail I missed before.
See https://github.com/kubernetes/kubernetes/blob/v1.26.3/staging/src/k8s.io/apiserver/pkg/registry/generic/registry/store.go#L568 and https://github.com/kubernetes/kubernetes/blob/v1.26.3/staging/src/k8s.io/apiserver/pkg/registry/generic/registry/store.go#L619

marshtompsxd:
@euclidgame This PR is ready to review

euclidgame (approved):

PR #213, opened 2023-08-14
Split the fluent controller into two reconcilers

marshtompsxd:
One reconciler is used for creating the secret object, which contains the configuration data of FluentBit, and the other one deploys FluentBit daemons that use the configuration data.


marshtompsxd:
@euclidgame could you also review this one?

PR #214, opened 2023-08-15
Complete and restructure the liveness proof of rabbitmq controller

euclidgame:
This PR proves the liveness property of the rabbitmq controller. Different from the previous proof, this time both the liveness property and the state machine specification become more complex: the liveness property includes the stateful set object managed by the controller, and the state machine specification has been enriched a lot with validation rules, garbage collector, finalizer, and many other features. Besides, the implementation of the rabbitmq controller has changed: it waits for the garbage collector to recycle the stateful set of the old custom resource to avoid accidental downscale (see #211 ). All these factors make it even more challenging to prove liveness.

### Key challenges

One of the key challenges during the proof is to identify and manage a large number of invariants. To give some concrete examples, here are two representative invariants:
1. If stateful set exists, then its owner reference must point to and only point to the current custom resource.
2. No delete request with the stateful set key will be sent to Kubernetes API.

The first is used to show that finally stateful set created from old custom resources (if any) will be deleted, and this is done by the garbage collector. The second is to verify that if we get the desired stateful set, it will always exist. The proof of 2 requires 1 because we need it to show the garbage collector never tries to delete the stateful set (the gc actions is never enabled).

Notably, though we call them "invariants", neither 1 nor 2 holds from the beginning of the execution (`init()`). That said, they are not the safety invariants that hold throughout the entire execution, but they start to hold at some point during the execution and continue to hold since then.

Proving the liveness property requires a large number of such invariants, so how to prove and use them becomes a significant part of the proof burden. Note that the same problem exists even before this PR.

### Proof structure

We notice that a common pattern in the liveness proof is that, to make the desired property eventually happens, the environment needs to first enter a state where there are no more perturbations that can interrupt the execution moving toward the desired property. For example, if we want the controller to make progress to the desired state, then we need to first prove that the controller stops crashing from a point.

That said, the entire liveness proof, similar to the actual execution, is divided into phases. Starting from the initial phase (the initial state of the state machine), the system first enters a phase where perturbations (e.g., staleness, crashes, disturbing requests) no longer exist, then moves to the final phase where the desired state is achieved by the controller's correct execution. At each phase, we derive some invariants (the properties that hold from this point) which will be used to prove the properties of the next phase. Naturally, the invariants used in the proof can be grouped by the phase. They were referred to as `assumptions`, `invariants_since_rest_id`, `invariants_led_to_by_rest_id` (these variable names are indeed confusing, and we stop calling them in this way).

In this proof, we divide the execution into different phase and invariants into several groups $I_1,I_2,\cdots,I_n$. More details in [this comment](https://github.com/vmware-research/verifiable-controllers/pull/214#issuecomment-1681302812).

In this structure, we often need to prove lemmas in the form of `spec |= true ~> []p`. If the property `p` talks about messages sent after a particular point, we often have to choose a `rest_id` and prove that eventually request messages before `req_id` are gone. This PR introduces a [lemma](https://github.com/vmware-research/verifiable-controllers/blob/f5783e1a014e7eedc76e38bfd82fa61de6d4cb98/src/kubernetes_cluster/proof/kubernetes_api_liveness.rs#L276C14-L276C75) in the `kubernetes_api_liveness` module to help reason about this kind of invariants, and the controller developers no longer need to directly reason about different `rest_id`s (And it's indeed better to avoid it). This improves both the reusability and the readability of the proof/lemmas.

This PR does not try to finish proving all the invariants but focuses on ensuring that:
1. The invariants are grouped correctly into phases that follow the same logical order of the proof/execution.
2. The invariants should be provable with all antecedent invariants required.

The phases of the invariants used in the liveness proof are:

- 0: Invariants that hold throughout the entire execution (i.e., safety invariants).
- 1: Crash and K8s busy are disabled, and the object that is in the schedule for reconciliation is updated.
- 2: The object being reconciled is updated.
- 3: Every creation and update request to the same object has the same effect.
- 4: Objects only have their owner references pointing to the current cr, not the previously-deleted cr.
- 5: No deletion requests are sent to the K8s API.


euclidgame:
I really like [this lemma](https://github.com/vmware-research/verifiable-controllers/blob/f5783e1a014e7eedc76e38bfd82fa61de6d4cb98/src/kubernetes_cluster/proof/kubernetes_api_liveness.rs#L276C14-L276C75) which shows that invalid api request messages will eventually all be consumed. I write an example using it here.

src/kubernetes_cluster/spec/builtin_controllers/garbage_collector.rs
```diff
@@ -37,11 +37,12 @@ pub open spec fn run_garbage_collector() -> BuiltinControllersAction {
...
             let resources = input.resources;
             let key = input.key;
-            let namespace = resources[key].metadata.namespace.get_Some_0();
+            let namespace = key.namespace;
             let owner_references = resources[key].metadata.owner_references.get_Some_0();
+            let n_owner_refs = owner_references.len();
             // The garbage collector decides whether to delete an object by checking its owner references,
             // it deletes the object if for each owner reference...
-            if forall |i| #![trigger owner_references[i]] {
+            if forall |i| #![trigger owner_references[i]] 0 <= i < n_owner_refs ==> {
```
marshtompsxd, line 45: Good catch.
Could you just replace `n_owner_refs` with `owner_references.len()`? (since you only use it once)

src/controller_examples/zookeeper_controller/proof/liveness/helper_invariants.rs
marshtompsxd, line 1: Are these invariants used for zk controller?

src/kubernetes_cluster/proof/kubernetes_api_liveness.rs
```diff
@@ -223,6 +246,176 @@ pub open spec fn no_req_before_rest_id_is_in_flight(rest_id: RestId) -> StatePre
     }
 }
 
+/// Newly created message means it doesn't exist in in_flight in last state but exists in the current state.
+pub open spec fn every_new_in_flight_req_msg_is_expected(expected_messages: FnSpec(Message) -> bool) -> ActionPred<Self> {
```
marshtompsxd, line 250: Maybe you can rename `_is_expected` to `_satisfies` and `expected_message` to `msg_pred`

src/controller_examples/rabbitmq_controller/exec/reconciler.rs
```diff
@@ -854,6 +853,29 @@ fn make_role_binding(rabbitmq: &RabbitmqCluster) -> (role_binding: RoleBinding)
...
+    owner_references.push(rabbitmq.controller_owner_ref());
+    proof {
+        assert_seqs_equal!(
+            owner_references@.map_values(|owner_ref: OwnerReference| owner_ref@),
+            rabbitmq_spec::make_role(rabbitmq@).metadata.owner_references.get_Some_0()
+        );
+    }
+    let mut metadata = found_stateful_set.metadata();
+    metadata.set_owner_references(owner_references);
+    metadata.reset_finalizers();
```
marshtompsxd, line 873: Can you add some comment on these set_owner_reference and reset_finalizers? Just to remind ourselves these are to make the proof go easier, without affecting the functionality.

src/controller_examples/rabbitmq_controller/proof/liveness/helper_invariants/invariants.rs
```diff
@@ -0,0 +1,863 @@
...
+    init_invariant(spec, RMQCluster::init(), RMQCluster::next(), invariant);
+}
+
+pub proof fn lemma_true_leads_to_always_every_update_cm_req_does_the_same(spec: TempPred<RMQCluster>, rabbitmq: RabbitmqClusterView)
+    requires
+        valid(stable(spec)),
+        spec.entails(always(lift_state(RMQCluster::every_in_flight_msg_has_lower_id_than_allocator()))),
+        spec.entails(always(lift_action(RMQCluster::next()))),
+        spec.entails(always(lift_state(RMQCluster::busy_disabled()))),
+        spec.entails(tla_forall(|i| RMQCluster::kubernetes_api_next().weak_fairness(i))),
```
marshtompsxd, line 477: Is this wf assumption used in this proof?

euclidgame: Of course. Otherwise req messages won't be consumed. concretely, it is used in the lemma (`lemma_true_leads_to_always_every_in_flight_req_msg_is_expected`) this lemma calls.

marshtompsxd: ic

src/controller_examples/rabbitmq_controller/proof/liveness/helper_invariants/invariants.rs
```diff
@@ -0,0 +1,863 @@
...
+        ),
+{
+    let requirements = |msg: Message| {
+        cm_update_request_msg(rabbitmq.object_ref())(msg)
+        ==> msg.content.get_update_request().obj.spec == ConfigMapView::marshal_spec((make_server_config_map(rabbitmq).data, ()))
+    };
+    eliminate_always(spec, lift_action(RMQCluster::next()));
+    eliminate_always(spec, lift_state(RMQCluster::each_key_in_reconcile_is_consistent_with_its_object()));
+    eliminate_always(spec, lift_state(RMQCluster::the_object_in_reconcile_has_spec_and_uid_as(rabbitmq)));
+    assert forall |ex| #[trigger] spec.satisfied_by(ex) implies lift_action(RMQCluster::every_new_in_flight_req_msg_is_expected(requirements)).satisfied_by(ex) by {
```
marshtompsxd, line 492: I feel like for this type of reasoning, we can have a rule similar to `init_invariant`, like:
```
pub proof fn init_invariant<T>(spec: TempPred<T>, init: StatePred<T>, next: ActionPred<T>, inv: ActionPred<T>)
    requires
        forall |s, s_prime: T| #[trigger] next(s, s_prime) ==> inv(s, s_prime),
        spec.entails(always(lift_action(next))),
    ensures
        spec.entails(always(lift_action(inv))),
```

euclidgame: Oh I feel like this doesn't sound better. Because will this need strengthen next before calling this lemma? The stronger next thing is a nightmare to me (stronger_next => entails_and ==> always_and ==> temp_equality).

marshtompsxd: That's fine. This is more like a personal taste/style on the proof. Strengthening next is indeed a super tedious process (unfortunately the macro is not strong enough). I am ok with that but I personally don't like exposing `ex` in the proof.

If you really prefer the current approach then you can keep it.

src/kubernetes_cluster/proof/kubernetes_api_liveness.rs
euclidgame: Done

src/controller_examples/rabbitmq_controller/exec/reconciler.rs
euclidgame: Done

src/controller_examples/rabbitmq_controller/proof/liveness/helper_invariants/owner_ref.rs
```diff
@@ -0,0 +1,297 @@
...
+    common::*,
+    proof::common::*,
+    spec::{rabbitmqcluster::*, reconciler::*},
+};
+use crate::temporal_logic::{defs::*, rules::*};
+use vstd::prelude::*;
+
+verus! {
+
+spec fn sts_key_not_exists_or_delete_msg_in_flight_or_sts_has_current_ref(rabbitmq: RabbitmqClusterView) -> StatePred<RMQCluster> {
```
marshtompsxd, line 27: Why not just say if key exists then either exists delete msg or owner ref points to the current cr?

src/controller_examples/rabbitmq_controller/proof/liveness/helper_invariants/owner_ref.rs
```diff
@@ -0,0 +1,297 @@
...
+        spec.entails(tla_forall(|i| RMQCluster::builtin_controllers_next().weak_fairness(i))),
+        spec.entails(always(lift_state(RMQCluster::desired_state_is(rabbitmq)))),
+        spec.entails(always(lift_state(server_config_map_has_no_finalizers_or_timestamp_and_only_has_controller_owner_ref(rabbitmq)))),
+        spec.entails(always(lift_state(every_update_cm_req_does_the_same(rabbitmq)))),
+        spec.entails(always(lift_state(every_create_cm_req_does_the_same(rabbitmq)))),
+    ensures
+        spec.entails(true_pred().leads_to(always(lift_state(server_config_map_has_owner_reference_pointing_to_current_cr(rabbitmq))))),
+{}
+
+pub proof fn lemma_eventually_only_valid_stateful_set_exists(spec: TempPred<RMQCluster>, rabbitmq: RabbitmqClusterView)
```
marshtompsxd, line 61: Could you write some comments for this lemma, since this is the first time we reason about gc? This lemma would be a very valuable reference for later proof.

marshtompsxd:
When I read the proof, sometimes it is a bit difficult to quickly figure out whether this invariants holds for the entire execution, or just after some point.
I think it would very helpful if we can easily distinguish the type 1 (`spec |= []invariant`) and type 2 invariants (`spec |= true ~> []invariant`) from the file path. Instead of just calling them helper_invariants, can we split them into two folders, one is `invariants` and one is `suffix_invariants` (feel free to use better names).

src/controller_examples/rabbitmq_controller/proof/liveness/helper_invariants/invariants.rs
```diff
@@ -0,0 +1,855 @@
...
+            &&& cm_create_request_msg(key)(msg)
+        } ==> {
+            &&& at_rabbitmq_step(key, RabbitmqReconcileStep::AfterCreateServerConfigMap)(s)
+            &&& msg == s.pending_req_of(key)
+            &&& s.network_state.in_flight.count(msg) == 1
+        }
+    }
+}
+
+#[verifier(external_body)]
```
marshtompsxd, line 157: Do you plan to prove these lemmas in this or the next PR?

src/temporal_logic/rules.rs
```diff
@@ -142,7 +142,7 @@ proof fn implies_apply_with_always<T>(ex: Execution<T>, p: TempPred<T>, q: TempP
     always_unfold::<T>(ex, p);
 }
 
-proof fn entails_apply<T>(ex: Execution<T>, p: TempPred<T>, q: TempPred<T>)
+pub proof fn entails_apply<T>(ex: Execution<T>, p: TempPred<T>, q: TempPred<T>)
```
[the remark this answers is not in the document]
marshtompsxd: As a side notes, I was planning to tear the rules.rs into several files focusing on different types of lemmas (e.g,. the ones directly talk about ex, the ones for quantifiers, wf1 ones, and so on) and use a prelude.rs to import the user-facing ones.

src/kubernetes_cluster/proof/kubernetes_api_liveness.rs
```diff
@@ -223,6 +246,182 @@ pub open spec fn no_req_before_rest_id_is_in_flight(rest_id: RestId) -> StatePre
...
+/// Note that in the precondition, we don't actually require always for this predicate, because the spec need to be stable and the always
+/// one can be derived. This also alleviate the preliminary work for the callers to use this lemma.
+///
+/// The last parameter must be equivalent to every_in_flight_req_msg_satisfies(requirements)
+pub proof fn lemma_true_leads_to_always_every_in_flight_req_msg_satisfies(
+    spec: TempPred<Self>, requirements: FnSpec(Message) -> bool, msg_state_pred: StatePred<Self>
+)
+    requires
+        valid(stable(spec)),
+        spec.entails(lift_action(Self::every_new_in_flight_req_msg_satisfies(requirements))),
```
marshtompsxd, line 285: I am trying to understand this precondition: what you later need is actually `spec.entails(always(lift_action(Self::every_new_in_flight_req_msg_satisfies(requirements))))` but `valid(stable(spec))` can easily give you that so you don't require the "always" here, is that right?

src/controller_examples/rabbitmq_controller/proof/liveness/helper_invariants/owner_ref.rs
euclidgame: I find it makes more sense to say `key_exists /\ owner_ref != [current_cr] ==> delete_msg_in_flight`.

src/kubernetes_cluster/proof/kubernetes_api_liveness.rs
euclidgame: Yes. We can also remove `valid(stable(spec))` and add `[]` to the second. If you want to use `strengthen_next_n`, the alternative is better.

src/temporal_logic/rules.rs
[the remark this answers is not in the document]
euclidgame: You'd better move later plans to the management file or somewhere else.

src/kubernetes_cluster/proof/kubernetes_api_liveness.rs
marshtompsxd: I also like this lemma a lot. The only counter-intuitive part is actually that this particular precondition does not have "always". I think it would be a lot easier to make sense of it if we just require the always here (and then we no longer need to require stable spec I guess?)

> This also alleviate the preliminary work for the callers to use this lemma.

I understand this but it seems to me proving the always is very simple (just one lemma call), especially if we later use the invariant_action lemma I was proposing.

src/controller_examples/rabbitmq_controller/proof/liveness/helper_invariants/owner_ref.rs
marshtompsxd: Then you can go that way

src/kubernetes_cluster/proof/kubernetes_api_liveness.rs
euclidgame: So I wrote this lemma before you proposed everything.... You have to combine all the always...

marshtompsxd: Yes, understand. I have some race conditions here. When I am writing the comment I don't always see your newest comments. If you plan to do the refactoring mentioned above pls go ahead, otherwise let's put it in the management file as you suggested above.

src/temporal_logic/rules.rs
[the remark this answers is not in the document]
marshtompsxd: yes

src/kubernetes_cluster/proof/kubernetes_api_liveness.rs
euclidgame: I made some changes (you can take a look). But I don't write the `invariant_action` because it's very obvious.

src/controller_examples/rabbitmq_controller/proof/liveness/helper_invariants/owner_ref.rs
euclidgame: I want to refactor this lemma to make it more general and move it to some common module. But I can add some temporary comments.

src/controller_examples/rabbitmq_controller/proof/liveness/liveness.rs
```diff
@@ -163,35 +208,96 @@ proof fn invariants_is_stable(rabbitmq: RabbitmqClusterView)
     );
 }
 
+spec fn invariants_phase_II(rabbitmq: RabbitmqClusterView) -> TempPred<RMQCluster> {
```
marshtompsxd, line 211: Could you document phase 2, 3, 4, 5, and 6? Just document intuitively what are these phases and why we want to split the entire proof process into such phases.

src/controller_examples/rabbitmq_controller/proof/liveness/liveness.rs
```diff
@@ -68,52 +86,75 @@ proof fn next_with_wf_is_stable()
...
+    .and(tla_forall(|input| RMQCluster::builtin_controllers_next().weak_fairness(input)))
+    .and(RMQCluster::disable_crash().weak_fairness(()))
+    .and(RMQCluster::disable_busy().weak_fairness(()))
+    .and(always(lift_state(RMQCluster::desired_state_is(rabbitmq))))
+    .and(invariants_phase_I(rabbitmq))
+    .and(invariants_phase_II(rabbitmq))
+    .and(invariants_phase_III(rabbitmq))
+    .and(invariants_phase_IV(rabbitmq))
+    .and(invariants_phase_V(rabbitmq))
+    .and(invariants_phase_VI(rabbitmq))
```
marshtompsxd, line 128: I like the idea of explicitly splitting the proof into phases (we've been doing that but here you make it very explicit) and group invariants into phases. I think finding the right way to split the execution into phases that are amenable to proof is one of the keys to liveness verification.

src/controller_examples/rabbitmq_controller/proof/liveness/liveness.rs
```diff
@@ -68,52 +86,75 @@ proof fn next_with_wf_is_stable()
...
+    next_with_wf_is_stable();
+    stable_and_temp(
+        next_with_wf(),
+        always(lift_state(RMQCluster::desired_state_is(rabbitmq)))
     );
 }
 
 // The safety invariants that are required to prove liveness.
-spec fn invariants(rabbitmq: RabbitmqClusterView) -> TempPred<RMQCluster> {
+spec fn invariants_phase_I(rabbitmq: RabbitmqClusterView) -> TempPred<RMQCluster> {
```
marshtompsxd, line 144: Could you make it explicit that these invariants hold from the init point?

src/controller_examples/rabbitmq_controller/proof/liveness/liveness.rs
euclidgame: That's what I talk about in the description

src/controller_examples/rabbitmq_controller/proof/liveness/liveness.rs
euclidgame: Actually I want to say Phase 0 but there is no zero in roman numerals 👎

marshtompsxd: You can say `phase_zero` if you don't mind the inconsistency

src/controller_examples/rabbitmq_controller/proof/liveness/liveness.rs
marshtompsxd: yes will do

src/controller_examples/rabbitmq_controller/proof/liveness/liveness.rs
```diff
@@ -163,35 +208,96 @@ proof fn invariants_is_stable(rabbitmq: RabbitmqClusterView)
     );
 }
 
+spec fn invariants_phase_II(rabbitmq: RabbitmqClusterView) -> TempPred<RMQCluster> {
+    always(lift_state(RMQCluster::crash_disabled()))
+    .and(always(lift_state(RMQCluster::busy_disabled())))
+    .and(always(lift_state(RMQCluster::the_object_in_schedule_has_spec_and_uid_as(rabbitmq))))
+}
+
+spec fn invariants_phase_III(rabbitmq: RabbitmqClusterView) -> TempPred<RMQCluster> {
```
marshtompsxd, line 217: Maybe you want to move this below `invariants_phase_II_is_stable`?

src/controller_examples/rabbitmq_controller/proof/liveness/liveness.rs
```diff
@@ -1894,4 +2803,45 @@ proof fn lemma_server_config_map_is_stable(
...
 
+proof fn lemma_stateful_set_is_stable(
+    spec: TempPred<RMQCluster>, rabbitmq: RabbitmqClusterView, p: TempPred<RMQCluster>
+)
+    requires
+        spec.entails(p.leads_to(lift_state(current_stateful_set_matches(rabbitmq)))),
+        spec.entails(always(lift_action(RMQCluster::next()))),
+        spec.entails(always(lift_state(helper_invariants::no_delete_sts_req_is_in_flight(rabbitmq.object_ref())))),
+        spec.entails(always(lift_state(helper_invariants::every_update_sts_req_does_the_same(rabbitmq)))),
+        spec.entails(always(lift_state(helper_invariants::stateful_set_has_no_finalizers_or_timestamp_and_only_has_controller_owner_ref(rabbitmq)))),
```
marshtompsxd, line 2814: IIUC this precondition is used to ensure k8s API does not delete the object when handling update requests (the delete-during-update path). If so, could you document this a bit (just to remind ourselves in the future)?

euclidgame:
After some thoughts, I find it's better to organize those invariants in this way:

- `invariants_all_the_time`: This is the combination of actions, weak fairness and those invariants held from beginning. As what I said when talking to @marshtompsxd , it's really beneficial to see those initial invariants same as next and weak fairness (which I will talk about later).
- Assumption: This is the assumption that custom resource always exists, i.e., `[]desired_state_is(cr)`.
- Other invariants sorted by phase.

We can describe the state of system in phase $N$ ($state_N$) as `invariants_all_the_time /\ assumption /\ invariants_phase_II /\ ... /\ invariants_phase_N` (for $N>1$) and `invariants_all_the_time /\ assumption` (for $N=1$). 

Suppose there are M phases. We first prove the liveness property under all invariants and assumption ( $state_M \models true \leadsto\Box matches(cr)$ ). Then the left goes to, for every $N < M$, we prove $state_N \models true\leadsto I_{N+1}$ ($I_i$ refers to `invariants_phase_i`). Then for every $N < M$, we get $state_N \models I_{N+1}\leadsto\Box matches(cr)$ by unpacking conditions from $state_{N+1}\models true\leadsto\Box matches(cr)$. So we have $state_N \models true\leadsto\Box matches(cr)$ for every $N$.

By unpacking $state_1$ we will have $init\wedge I_1\models assumption\leadsto\Box matches(cr)$ ($I_1$ means `invariants_all_the_time`). Replace $init\wedge I_1$ with `cluster_spec()` because these two things are obviously equivalent (We derive this by equality, not the `simplify_predicate` in current proof).

If we name other invariants in another name (say, xxxxx), then our handling of xxxxx of different phase will be all the same. And the proof will be more systematic.

src/controller_examples/rabbitmq_controller/proof/liveness/liveness.rs
euclidgame: The precondition is an umbrella of several requirements. No timestamp is enough for showing that the object won't be deleted during update; no finalizers is added because this will make it easier to prove no timestamp without adding extra invariants; only having controller owner ref seems useless here, but it's just more convenient to put these invariants together and prove them all at once.

marshtompsxd:
@euclidgame 
I like your proposal, except that for `invariants_all_the_time` you can just call them `invariants`, and for `invariants_phase_i` you can probably call them `invariants_since_phase_i`.
Calling `next` and `wf` as `invariants` is not a very common practice (because they are not derived), but if you think this can bring a lot of benefits then go ahead.

marshtompsxd:
btw, I reworded the description of this PR. Please take a read and make sure I didn't mis-word anything. You can add the comments to the phases as we discussed. If the proof structure changes, please also update the description.

src/controller_examples/rabbitmq_controller/proof/liveness/helper_invariants/owner_ref.rs
marshtompsxd: Looking forward

euclidgame:
> Calling next and wf as invariants is not a very common practice (because they are not derived), but if you think this can bring a lot of benefits then go ahead.

I can avoid calling them invariants. I reword that comment, and I think I make it clearer.

src/controller_examples/rabbitmq_controller/proof/liveness/liveness.rs
euclidgame: Added some comments.

src/temporal_logic/rules.rs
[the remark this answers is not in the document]
marshtompsxd: I guess this one no longer needs to be pub?

src/temporal_logic/rules.rs
```diff
@@ -1126,6 +1147,38 @@ macro_rules! strengthen_next_n_internal {
...
+///     spec |= []p2
+///         ...
+///     spec |= []pn
+///     next == p1 /\ p2 /\ ... /\ pn
+/// post:
+///     spec |= []inv
+/// 
+/// Usage: invariant_action_n!(spec, next, inv, p1, p2, ..., pn)
+#[macro_export]
+macro_rules! invariant_action_n {
```
marshtompsxd, line 1164: I was about to say that we should make the macros consistent with `strengthen_next_n` on the order of arguments, but I also realized that there is no need to differentiate `next` and other `p` in `strengthen_next_n`. I think putting `spec` before `next` (actually the strengthened next) and treating all others as `px` is the right way to go. I will refactor `strengthen_next_n` in this way later.

src/kubernetes_cluster/proof/kubernetes_api_liveness.rs
```diff
@@ -223,6 +246,209 @@ pub open spec fn no_req_before_rest_id_is_in_flight(rest_id: RestId) -> StatePre
...
+/// Previously, when "requirements" was irrelavant to the state, b will be sure to hold. Later, we find that "requirements" in some 
+/// case does need some information in the state. So we add state as another parameter and requires the caller of the lemma
+/// lemma_true_leads_to_always_every_in_flight_req_msg_satisfies to prove b. is always satisfied. In order not to make those cases
+/// where "requirements" has nothing to do with state more difficult, we combine a. and b. together.
+/// 
+/// Therefore, we have the following predicate. As is easy to see, this is similar as:
+///     (s.in_flight(msg) ==> requirements(msg, s)) ==> (s_prime.in_flight(msg) ==> requirements(msg, s_prime))
+/// If we think of s.in_flight(msg) ==> requirements(msg, s) as an invariant, it is the same as the proof of invariants in previous
+/// proof strategy.
+pub open spec fn every_new_req_msg_if_in_flight_then_always_satisfies(requirements: FnSpec(Message, Self) -> bool) -> ActionPred<Self> {
```
marshtompsxd, line 267: Shall we remove `always` from the name?

src/kubernetes_cluster/proof/kubernetes_api_liveness.rs
```diff
@@ -223,6 +246,209 @@ pub open spec fn no_req_before_rest_id_is_in_flight(rest_id: RestId) -> StatePre
...
+}
+
+/// This lemma shows that if spec ensures every newly created Kubernetes api request message satisfies some requirements,
+/// the system will eventually reaches a state where all Kubernetes api request messages satisfy those requirements.
+///
+/// To require "every newly create Kubernetes api request message satisfies some requirements", we use a FnSpec (i.e., a closure)
+/// as parameter which can be defined by callers and require spec |= [](every_new_req_msg_if_in_flight_then_always_satisfies(requirements)).
+///
+/// The last parameter must be equivalent to every_in_flight_req_msg_satisfies(requirements)
+pub proof fn lemma_true_leads_to_always_every_in_flight_req_msg_satisfies(
```
marshtompsxd, line 295: Actually, I would recommend keeping this GREAT lemma even simpler by removing the last parameter from this lemma and just use `Self::every_in_flight_req_msg_satisfies(requirements)` in the postcondition. If the caller needs to convert it to a msg_state_pred, the caller should call `temp_pred_equality`. This lemma should just focus on proving `true ~> []every_in_flight_req_msg_satisfies(requirements)`.

src/kubernetes_cluster/proof/kubernetes_api_liveness.rs
```diff
@@ -223,6 +246,209 @@ pub open spec fn no_req_before_rest_id_is_in_flight(rest_id: RestId) -> StatePre
...
+    );
+
+    // Then unpack the condition from spec.
+    unpack_conditions_from_spec(
+        stable_spec, lift_state(Self::rest_id_counter_is(rest_id)), true_pred(),
+        always(lift_state(Self::every_in_flight_req_msg_satisfies(requirements)))
+    );
+    temp_pred_equality(true_pred().and(lift_state(Self::rest_id_counter_is(rest_id))), lift_state(Self::rest_id_counter_is(rest_id)));
+    entails_trans(spec, stable_spec, lift_state(Self::rest_id_counter_is(rest_id)).leads_to(always(lift_state(Self::every_in_flight_req_msg_satisfies(requirements)))));
+}
```
marshtompsxd, line 451: Can we leave one empty line before the next lemma :-))

src/controller_examples/rabbitmq_controller/proof/liveness/liveness.rs
```diff
@@ -50,12 +67,23 @@ proof fn liveness_proof_forall_rabbitmq()
...
+    .and(always(lift_state(RMQCluster::desired_state_is(rabbitmq))))
+    .and(invariants_since_phase_I(rabbitmq))
+    .and(invariants_since_phase_II(rabbitmq))
+    .and(invariants_since_phase_III(rabbitmq))
+    .and(invariants_since_phase_IV(rabbitmq))
+    .and(invariants_since_phase_V(rabbitmq))
+}
+
 // Next and all the wf conditions.
 spec fn next_with_wf() -> TempPred<RMQCluster> {
```
marshtompsxd, line 81: We probably should just move it to the cluster state machine and replace that `partial_spec` (if that still exists).

euclidgame:
I really don't like the `at_most_one_xxx_req_xxxxx` invariants. They say too much that may not be useful, and they take too long to finish. @marshtompsxd , you'd better revisit them and try to come up with some alternatives. If I have time, I will too.

marshtompsxd:
For```
    |s: RMQCluster| {
        forall |msg| {
            &&& #[trigger] s.network_state.in_flight.contains(msg)
            &&& sts_create_request_msg(key)(msg)
        } ==> {
            &&& at_rabbitmq_step(key, RabbitmqReconcileStep::AfterCreateStatefulSet)(s)
            &&& msg == s.pending_req_of(key)
            &&& s.network_state.in_flight.count(msg) == 1
        }
    }
```
I think we can take `s.network_state.in_flight.count(msg) == 1` out of it, which hopefully makes the verification faster. And we can have a separate safety invariant saying that for each msg in the network, its count is 1 (which is obvious because of the rest_id).

File: src/controller_examples/rabbitmq_controller/exec/reconciler.rs [hunk 21]
```diff
@@ -576,6 +573,35 @@ fn make_plugins_config_map(rabbitmq: &RabbitmqCluster) -> (config_map: ConfigMap
...
+    }
+    let mut metadata = found_config_map.metadata();
+
+    // Since we requirement the owner_reference only contains current cr, this set operation won't change anything.
+    // Similarly, we never set finalizers for any stateful set, resetting finalizers won't change anything.
+    // The reason why we add these two operations is that it makes the proof easier.
+    // In this way, we can easily show that what the owner references and finalizers of the object in every update request 
+    // for stateful set are.
+    metadata.set_owner_references(owner_references);
+    metadata.reset_finalizers();
```
marshtompsxd, line 599: Can you rename it to `unset_finalizers`?

src/controller_examples/rabbitmq_controller/proof/liveness/helper_invariants/invariants.rs
```diff
@@ -0,0 +1,1215 @@
...
+    }
+}
+
+// TODO: fix this lemma.
+// If the configmap has no owner_reference, fixing the lemma is simple:
+// we just need to show that the configmap in the cluster state never has any owner reference.
+//
+// However, later we are going to set the owner_reference of the configmap to the CR object,
+// so we will need the assumption that "CR always exists" to prove this invariant.
+pub proof fn lemma_true_leads_to_always_no_delete_cm_req_is_in_flight(spec: TempPred<RMQCluster>, rabbitmq: RabbitmqClusterView)
```
marshtompsxd, line 551: You can remove the comments now

marshtompsxd:
@euclidgame You can double-check the description and see if there is anything you want to update. I see there are still two invariant lemmas marked as external, and I assume you plan to prove them in the next PR. I will merge it after the CI passes.

PR #217, opened 2023-08-18
Fix some dormant bugs in the zookeeper controller

marshtompsxd:
There were some dormant liveness and safety bugs in the zookeeper controller's implementation and this PR focuses on fixing them. Besides that, this PR also does the following:
- Simplifying the control flow of the zookeeper controller a bit: there is no need to always set the zk node at the end of each reconciliation. Since the data in the zk node is only needed during downscaling, we only need to set the data right before updating the stateful set.
- Revising the deploy script to make it easier to test a local version of the controller instead of always running the CI to build the images and pulling from the GitHub registry.
- Using a more compact model to specify the zk API. We will likely need a more detailed specification with multiple actions such as setting the parent node and querying the version number to better reason about its correctness, as described below.

We described the details of the bugs as follows.

### Liveness bug caused by an intermediate state
The controller will never successfully downscale the stateful set if it crashes right after creating the parent zk node (`/zookeeper-operator`) and before creating the child zk node (right [here](https://github.com/vmware-research/verifiable-controllers/pull/217/files#diff-29edc4ca79b3dde9721591f2532c6d500db3b859ad4132f0d07c5a521aaecb90L72)). After restart, the controller will try to look up the child zk node again, and since it does not exist it tries to create the parent node again. The creation will deterministically return the error `ZkError::NodeExists` because the parent node already exists, and the controller will exit reconciliation early because of the error. As a result, the controller will never be able to create the child zk node, let alone update the stateful set. This is a typical intermediate state bug example: the controller cannot start over from the intermediate state where the parent node exists but the child node does not.

The solution is to proceed to create the child zk node even if the parent node already exists, as [here](https://github.com/vmware-research/verifiable-controllers/pull/217/files#diff-90b2df410ce4d4305bf6b82eb7046e10bb94649b6de831bddfd59c4787ea378dR111).

### Safety bug caused by missing error check
The controller will continue to issue scaling operations even if it fails to set the zk node. This gives no guarantee that the data in the zk node is consistent with the desired `replicas` when downscaling the stateful set.

The solution is to check the result of setting the zk node and only proceed to update the stateful set if it is OK, as [here](https://github.com/vmware-research/verifiable-controllers/pull/217/files#diff-190b43e4f61437860f3ca3e4deff0d458de97589eb538f058c140cb4b0999c93R230).

### Safety bug caused by a race condition
The controller might issue unsafe scaling operations if it crashes right after sending the request to set the data of the child zk node, meanwhile, the user changes the `replicas` number in the custom resource object. Consider an execution starting from where the current replica number of the zookeeper cluster is 5 and the newly desired `replicas` is 4. The controller sends a request to set the data of the zk node to "4" and then crashes (the set request is not handled yet). Meanwhile, the user changes the desired `replicas` to 3. After restart, this time the controller sets the data to "3". Right after the controller sets the data to "3" and before the controller updates the stateful set, the previous request (that tries to set the data to "4") gets handled by the zookeeper cluster so the data is overwritten to "4". Then the controller updates the stateful set which triggers downscale. During the downscale, both the 4th and 5th zookeeper nodes should remove from the membership list but it turns out that the 4th will remain in the membership list because the zk node shows "4", instead of "3".

The solution is to check the version number when setting the zk node [here](https://github.com/vmware-research/verifiable-controllers/pull/217/files#diff-90b2df410ce4d4305bf6b82eb7046e10bb94649b6de831bddfd59c4787ea378dR122). The version reflects how many times this zk node gets updated, and the set operation will fail if the provided version does not match the actual version of the zk node. This can be used to prevent unexpected race conditions as described above: The late set request (with "4") won't succeed because its version number is invalidated by the previous set request (with "3").

The three bugs are discovered during the process of formally reasoning about the liveness and safety of the controller, and we further need a more detailed specification of the zk API and a machine-checkable proof to ensure the bugs are correctly fixed and there are no more such bugs.


euclidgame:
Actually I noticed the first two before but I thought it was a temporary implementation so didn't pay too much attention. For the third bug, what determines the order in which Zookeeper processes requests? Actually I don't understand why the version number prevents the bad behavior. It seems to me that the version number is fetched from the existing node. If the "3" update request has already been handled, won't the version number be changed as well?

marshtompsxd:
> Actually I noticed the first two before but I thought it was a temporary implementation so didn't pay too much attention.

Please report such bugs next time if you notice any.

> For the third bug, what determines the order in which Zookeeper processes requests? Actually I don't understand why the version number prevents the bad behavior. It seems to me that the version number is fetched from the existing node. If the "3" update request has already been handled, won't the version number be changed as well?

If the "3" update is handled, then the version number increments, and then it becomes larger than the version number carried by the "4" update, so the pending "4" update will never succeed as long as "3" succeeds.

euclidgame:
I see. What you worries about is that the previous request has already been sent. I thought you were worrying about two `set_zk_node_internal` processes happen concurrently.

euclidgame (approved):

PR #220, opened 2023-08-22
Prove safety property on scaling-down for Rabbitmq controller

euclidgame:
Prove the property that stateful set never scales down in rabbitmq controller. Details can be found in #223 . To supplement is, the concrete invariants in the code are as follows:

- Invariant 1: `stateful_set_not_scaled_down`
- Invariant 2: `object_in_sts_update_request_has_smaller_rv_than_etcd`
- Invariant 3: `replicas_of_stateful_set_update_request_msg_is_no_smaller_than_etcd`
- Invariant 4: `replicas_of_etcd_stateful_set_satisfies_order`
- Invariant 5: `object_in_ok_get_response_has_smaller_rv_than_etcd`
- Invariant 6: `response_at_after_get_stateful_set_step_is_sts_get_response`
- Invariant 7: `object_in_ok_get_resp_is_same_as_etcd_with_same_rv`
- Invariant 7: `replicas_of_stateful_set_create_or_update_request_msg_satisfies_order`
- Invariant 8: `etcd_and_scheduled_and_triggering_cr_in_correct_order`
- Invariant 9: `object_in_every_create_or_update_request_msg_only_has_valid_owner_references`
- Invariant 10: `every_owner_ref_of_every_object_in_etcd_has_different_uid_from_uid_counter`

Each corresponds to a spec function.

PR #221, opened 2023-08-22
Let the client state machine choose the input

marshtompsxd:
We don't need wf assumptions for the client state machine, so we should just let the client state machine to choose the input (from its step).


marshtompsxd:
@euclidgame Could you review this one before working on #220 ?

euclidgame (approved):

PR #224, opened 2023-08-26
Use simple and atomic Zookeeper API specification

marshtompsxd:
The new specification describes the behavior of three Zookeeper API `exists`, `create` and `set_data`:
- `exists` looks up the existence of a ZK node
- `create` creates a new ZK node with the associated data only when its parent node already exists
- `set_data` updates the data of an existing ZK node when the provided version number matches

A new shim layer for calling the Zookeeper API comes with the new specification. Compared to the previous one, the new one is atomic (as the specification) and simply wraps around one Zookeeper client API: the shim layer sets up the connection to Zookeeper, prepares the argument, calls only *one* Zookeeper API, and returns the result back to `recocile_core`.


src/controller_examples/zookeeper_controller/spec/zookeeper_api.rs
```diff
@@ -8,92 +8,208 @@ use vstd::{prelude::*, string::*};
...
+}
+
+// handle_set_data models the behavior of the zookeeper server handling the set data request.
+// To set the data, the node needs to exist and the provided version number must match the current version of the node.
+pub open spec fn handle_set_data(
+    name: StringView, namespace: StringView, path: Seq<StringView>, data: ZKNodeValue, version: ZKNodeVersion, resources: StoredState, state: ZKState
+) -> (ZKState, ZKAPISetDataResultView) {
+    let key = ObjectRef { kind: Kind::StatefulSetKind, namespace: namespace, name: name };
+    if !validate(name, namespace, path, resources) {
+        (state, ZKAPISetDataResultView{res: Err(Error::ZKNodeCreateFailed)})
```
euclidgame, line 195: Why is here `ZKNodeCreateFailed`?

src/controller_examples/zookeeper_controller/spec/zookeeper_api.rs
```diff
@@ -8,92 +8,208 @@ use vstd::{prelude::*, string::*};
...
+    }
+}
+
+// handle_create models the behavior of the zookeeper server handling the create request.
+// The creation succeeds only when (1) the node does not exist yet and (2) the parent node exists.
+pub open spec fn handle_create(
+    name: StringView, namespace: StringView, path: Seq<StringView>, data: ZKNodeValue, resources: StoredState, state: ZKState
+) -> (ZKState, ZKAPICreateResultView) {
+    let key = ObjectRef { kind: Kind::StatefulSetKind, namespace: namespace, name: name };
+    if !validate(name, namespace, path, resources) {
```
euclidgame, line 168: Actually I am a bit confused. Is it true that when we are here, the node must exist?

src/controller_examples/zookeeper_controller/spec/zookeeper_api.rs
marshtompsxd: Good catch. CP bug again

src/controller_examples/zookeeper_controller/spec/zookeeper_api.rs
marshtompsxd: Which node must exist? You mean the parent node?

src/controller_examples/zookeeper_controller/spec/zookeeper_api.rs
```diff
@@ -8,92 +8,208 @@ use vstd::{prelude::*, string::*};
...
+// Note that it uses the uid to avoid querying the data belonging to the old stateful set object.
+pub open spec fn handle_exists(
+    name: StringView, namespace: StringView, path: Seq<StringView>, resources: StoredState, state: ZKState
+) -> (ZKState, ZKAPIExistsResultView) {
+    let key = ObjectRef { kind: Kind::StatefulSetKind, namespace: namespace, name: name };
+    if !validate(name, namespace, path, resources) {
+        (state, ZKAPIExistsResultView{res: Err(Error::ZKNodeExistsFailed)})
+    } else {
+        let current_uid = resources[key].metadata.uid.get_Some_0();
+        let addr = ZKNodeAddr::new(name, namespace, current_uid, path);
```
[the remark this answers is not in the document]
marshtompsxd: I tried this but I found it does not really save code or make things more clear in the spec, so I keep it as it is.

src/controller_examples/zookeeper_controller/spec/zookeeper_api.rs
```diff
@@ -8,92 +8,205 @@ use vstd::{prelude::*, string::*};
...
+// handle_set_data models the behavior of the zookeeper server handling the set data request.
+// To set the data, the node needs to exist and the provided version number must match the current version of the node.
+pub open spec fn handle_set_data(
+    name: StringView, namespace: StringView, path: Seq<StringView>, data: ZKNodeValue, version: ZKNodeVersion, resources: StoredState, state: ZKState
+) -> (ZKState, ZKAPISetDataResultView) {
+    let key = ObjectRef { kind: Kind::StatefulSetKind, namespace: namespace, name: name };
+    if !validate(name, namespace, path, resources) {
+        (state, ZKAPISetDataResultView{res: Err(Error::ZKNodeSetDataFailed)})
+    } else {
+        let addr = ZKNodeAddr::new(name, namespace, resources[key].metadata.uid.get_Some_0(), path);
```
euclidgame, line 195: Sorry, I mean in this case, the node should already exist?

marshtompsxd: Yes: https://github.com/bonifaido/rust-zookeeper/blob/master/src/zookeeper.rs#L472-L486

euclidgame: So why you still check the existence here?

marshtompsxd: If we don't check, how do we know whether the node exists? Just to clarify this spec function is used to model the behavior of how zookeeper server handles the set_data request.

marshtompsxd: In other words, this spec function should precisely specify this https://github.com/bonifaido/rust-zookeeper/blob/master/src/zookeeper.rs#L472-L486

euclidgame: Because when you call this function, the node must exist.

marshtompsxd: By "you", do you mean any possible user, or just our zk controller?

PR #225, opened 2023-08-28
Add validation for finalizers

marshtompsxd:
The new validation rule states that if the object is marked to be deleted (`.metadata.deletion_timestamp.is_Some()`), then one cannot update the object by adding more finalizers to it.

### DIFF
--- a/src/controller_examples/rabbitmq_controller/proof/safety/safety.rs
+++ b/src/controller_examples/rabbitmq_controller/proof/safety/safety.rs
@@ -22,11 +22,11 @@ use vstd::prelude::*;
 verus! {
 
 /// To prove the safety property about stateful set, we need to first specify what the property is.
-/// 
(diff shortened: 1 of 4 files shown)


euclidgame (approved):

src/controller_examples/rabbitmq_controller/proof/safety/safety.rs
```diff
@@ -22,11 +22,11 @@ use vstd::prelude::*;
 verus! {
 
 /// To prove the safety property about stateful set, we need to first specify what the property is.
-///
```
euclidgame, line 25: You have been busy with fixing these....

marshtompsxd: No it is actually automatically fixed by my formatter. Not intended.

PR #233, opened 2023-09-05
Fix wrong key in user secret

marshtompsxd:


marshtompsxd:
@euclidgame Can you help me double-check whether after the fix the user secret is correct? Basically, run the official rabbitmq operator and check its content in the user secret to see if our key is correct.

PR #272, opened 2023-09-16
Support version upgrading for rabbitmq and enhance e2e tests

euclidgame:


src/deps_hack/src/lib.rs
```diff
@@ -107,6 +107,9 @@ pub struct ZookeeperConfig {
 #[kube(shortname = "rbmq", namespaced)]
 pub struct RabbitmqClusterSpec {
     pub replicas: i32,
+    /// Image is the name of the RabbitMQ docker image to use for RabbitMQ nodes in the RabbitmqCluster.
+    /// Must be provided together with ImagePullSecrets in order to use an image in a private registry.
+    pub image: String,
```
marshtompsxd, line 112: Did you copy the description from the official rabbitmq controller? We don't have an ImagePullSecrets here.

src/controller_examples/rabbitmq_controller/spec/reconciler.rs
```diff
@@ -778,7 +778,13 @@ pub open spec fn make_rabbitmq_pod_spec(rabbitmq: RabbitmqClusterView) -> PodSpe
             seq![
                 ContainerView::default()
                 .set_name(new_strlit("setup-container")@)
-                .set_image(new_strlit("rabbitmq:3.11.10-management")@)
+                .set_image(rabbitmq.spec.image)
+                .set_resources(
```
marshtompsxd, line 782: This reminds me: we should have a `resources` field for rabbitmq, like zookeeper

marshtompsxd:
@euclidgame Can you also add the e2e test to test version upgrading for rabbitmq controller in the same PR?

src/controller_examples/rabbitmq_controller/spec/reconciler.rs
euclidgame: We already have one..

marshtompsxd: just realize this is for the init_container, then nvm

tianyin:
@marshtompsxd Is this caused by flakiness due to the timing?

marshtompsxd:
I suspect so. I add a bit more waiting time before the failed test in another PR to hopefully remediate it.

PR #283, opened 2023-09-19
Add labels and annotations to rabbitmq

euclidgame:


src/controller_examples/rabbitmq_controller/exec/reconciler.rs
```diff
@@ -402,6 +403,18 @@ pub fn reconcile_core(rabbitmq: &RabbitmqCluster, resp_o: Option<Response<EmptyT
...
 }
 
+fn make_labels(rabbitmq: &RabbitmqCluster) -> (labels: StringMap)
+    requires
+        rabbitmq@.metadata.name.is_Some(),
+    ensures
+        labels@ == rabbitmq_spec::make_labels(rabbitmq@),
+{
+    let mut labels = StringMap::empty();
+    labels.insert(new_strlit("app").to_string(), rabbitmq.name().unwrap());
```
marshtompsxd, line 413: If you want every label to have this app=rabbitmq_name, you might want to switch the order of extend here: the argument of extend will overwrite whatever in self

src/controller_examples/rabbitmq_controller/exec/reconciler.rs
```diff
@@ -615,27 +619,38 @@ requires
...
-    found_config_map
+    config_map.set_data({
+        let old_data_opt = found_config_map.data();
+        let mut old_data = if old_data_opt.is_some() { old_data_opt.unwrap() } else { StringMap::empty() };
+        old_data.extend(made_server_cm.data().unwrap());
+        old_data
+    });
+    config_map.set_metadata({
+        let mut metadata = found_config_map.metadata();
+        // Since we requirement the owner_reference only contains current cr, this set operation won't change anything.
```
marshtompsxd, line 633: I thought we only check owner reference for statefulset and return if the owner reference does not match. Do we do the same for the configmap?

euclidgame: Yes. The invariants we need for config map are the same as stateful set.

marshtompsxd: Can you remind me why we want to stop reconcile and return error when the found configmap has unmatching owner reference? I know we did so for statefulset to prevent that tricky bug.

euclidgame: I don't think we have this. Pleas make sure you get it right.

euclidgame: The comment here was copied from stateful set, so there should be some mistakes.

marshtompsxd: But you answered yes to my first question, which was asking whether the controller checks owner reference for configmap.

euclidgame: Yeah I realized that. Because this code block is for making config map, so I got the question wrong.

marshtompsxd: So I guess the comment is wrong (since the configmap here might have an old owner reference)?

euclidgame: The comment is wrong, but we did decide to set the owner references for config map.

euclidgame: I think I can come up with a way to avoid setting the owner_references for updates.

src/controller_examples/rabbitmq_controller/exec/reconciler.rs
```diff
@@ -682,6 +691,14 @@ fn make_server_config_map(rabbitmq: &RabbitmqCluster) -> (config_map: ConfigMap)
         }
         rmq_conf_buff
     });
+    if rabbitmq.spec().rabbitmq_config().is_some() && rabbitmq.spec().rabbitmq_config().unwrap().advanced_config().is_some() 
+    && !rabbitmq.spec().rabbitmq_config().unwrap().advanced_config().unwrap().eq(&new_strlit("").to_string()) {
+        data.insert(new_strlit("advanced.config").to_string(), rabbitmq.spec().rabbitmq_config().unwrap().advanced_config().unwrap());
```
marshtompsxd, line 696: So we have the two fields `advanced_config` and `env_config` in the cr but we didn't use them before?

euclidgame: We didn't :-)

src/controller_examples/rabbitmq_controller/exec/reconciler.rs
```diff
@@ -732,11 +749,8 @@ fn make_service_account(rabbitmq: &RabbitmqCluster) -> (service_account: Service
             }
             owner_references
         });
-        metadata.set_labels({
-            let mut labels = StringMap::empty();
-            labels.insert(new_strlit("app").to_string(), rabbitmq.name().unwrap());
-            labels
-        });
+        metadata.set_labels(make_labels(rabbitmq));
```
marshtompsxd, line 752: @euclidgame It seems to me you never update the service account, role and role binding. In that case you might not set the labels and annotations to them since the two fields are not immutable. Otherwise you will have difficulty in writing liveness property for them (their labels/annotations won't be consistent with your cr)

euclidgame: We should update those objects later.

src/controller_examples/rabbitmq_controller/spec/reconciler.rs
```diff
@@ -523,26 +543,46 @@ pub open spec fn make_server_config_map(rabbitmq: RabbitmqClusterView) -> Config
...
+        },
+        data: Some({
+            let data = Map::empty()
+                        .insert(new_strlit("operatorDefaults.conf")@, default_rbmq_config(rabbitmq))
+                        .insert(new_strlit("userDefinedConfiguration.conf")@,
+                        {
+                            if rabbitmq.spec.rabbitmq_config.is_Some()
+                            && rabbitmq.spec.rabbitmq_config.get_Some_0().additional_config.is_Some()
+                            {   // check if there are rabbitmq-related customized configurations
+                                new_strlit("total_memory_available_override_value = 1717986919\n")@ + rabbitmq.spec.rabbitmq_config.get_Some_0().additional_config.get_Some_0()
```
marshtompsxd, line 563: Interesting. Is this also hardcoded in the official controller? And if so, do you know why?

euclidgame: Actually in the official controller, this key-value is inserted only when the cr specifies the memory limit (i.e., `cr.spec.resources.limits["memory"]`) and the limit will be the value. So it's conditional. 

But @ZichengMa set a default value here without any conditions. I never checked this before.

euclidgame: The `limits` field of `ResourceRequirements` is added by me and also the `resources` field of cr was added recently, so we must not have had that memory limit before.

euclidgame: Sure

marshtompsxd: Just a note: to implement that feature you might need to call StringMap.get which is currently `pub fn get<'a>(&'a self, key: String) -> (v: Option<StrSlice<'a>>)` -- if this one does not work well for your case pls feel free to implement other getter method that directly returns Option<String>

src/controller_examples/rabbitmq_controller/exec/reconciler.rs
marshtompsxd: I remember that this decision is to make the proof easier (when proving invariants). I don't have a strong opinion on this so you can decide what is the best way.

src/controller_examples/rabbitmq_controller/spec/reconciler.rs
euclidgame: Another thing I want to ask is how to compare a `Quantity` value to zero? Should I write a parser function?

euclidgame: Not only they require there is a memory limit in the resource field, but also the limit should not be zero.

marshtompsxd: I think you can skip that limit !=0 check for now -- they are trying to ensure they put in a correct limit, but honestly speaking only checking whether it equals zero is far from enough -- e.g., 1KB is not zero, but does not sound like a correct limit. BTW, do they directly use the Quantity literal as `total_memory_available_override_value` or they need to covert it somehow (like the unites might be different)?

euclidgame: They do a series of transitions which look non-trivial.

euclidgame: But the methods they use are all provided by the `quantity` module of k8s api

PR #292, opened 2023-09-21
Add steps to update all objects for rabbitmq controller

euclidgame:
Complete the reconcile process for rabbitmq controller by adding the response check for every step that is expected to receive a response and supporting updating all objects created by the controller.

I have planned to refactor the reconcile process to make it more concise. But let's keep it as now before we make sure how to avoid making a big influence on the proof.


e2e/src/rabbitmq_e2e.rs
```diff
@@ -625,3 +645,57 @@ pub async fn rabbitmq_e2e_test() -> Result<(), Error> {
     println!("E2e test passed.");
```
marshtompsxd, line 645: You can probably remove scaling_test() from this one since you already have a rabbitmq_scaling_e2e_test() below

euclidgame: sorry just forgot

src/kubernetes_api_objects/role.rs
```diff
@@ -52,6 +52,17 @@ impl Role {
         self.inner.metadata = metadata.into_kube();
     }
 
+    #[verifier(external_body)]
+    pub fn take_policy_rules(&mut self, other_role: Role)
```
marshtompsxd, line 56: I think this one does not have to be external_body. You should be able to implement it with a get and a set.

marshtompsxd: It is a minor thing. We can deal with this later.

PR #297, opened 2023-09-22
Add wrappers for EnvVar-related types

euclidgame:


reference-controllers/rabbitmq-controller/src/statefulset.rs
```diff
@@ -334,7 +334,7 @@ fn setup_container(rabbitmq: &RabbitmqCluster) -> corev1::Container{
     if rabbitmq.spec.image.is_some() {
         image_used = rabbitmq.spec.image.clone();
     }
-    let setup_container = corev1::Container {
+    let setup_container = corev1:: {
```
marshtompsxd, line 337: Maybe you can restore this change, tho we don't need this reference controller

PR #300, opened 2023-09-22
Deduplicate rabbitmq controller code using `ResourceBuilder`

euclidgame:
This PR aims to make the implementation of rabbitmq controller better arranged. 

The first thing we do is to make the reconcile step type (`RabbitmqReconcileStep`) more expressive: instead of literally concatenating the request type and resource type, we split them and use two fields for the step. The program can't tell what the previous step type represents so we have to exhaustively handle all cases which introduces many duplications. But now we can talk about what the request type is and what the resource object type created/updated/got is.

Then we can independently handle the creation/update process of each kind of object in their builder, which means each type of resource object will have its owner `ResourceBuilder`. Ideally, `ResourceBuilder` should be able to:

- Receive a cr and an old object (of `DynamicObject` type) and return an updated object or error;
- Receive a cr and return a brand new object or error.
- Return the basic information of the object: name, namespace, api_version, etc.
Those methods all handle dynamic objects instead of the unmarshaled resource object so that: (1) we don't have need an extra get result checker (and actually the get result checker is for update) and can simply do it checking together with the check before update; (2) it reduced the layers of checking results when processing k8s get response; (3) we don't need to marshal the updated/created object in the reconciler which is not convenient.

What the next step is should be bound to the step and decided by the reconciler; put it in another word, the resource builder should only care about the very object and should not be aware of what the next object to process or the subsequent reconcile state. But currently we have a method `state_after_create_or_update` which is against such a principle. We may refine it later.


src/controller_examples/rabbitmq_controller/spec/resource/common.rs
marshtompsxd, line 1: We can probably make it generic to all controllers later

src/controller_examples/rabbitmq_controller/spec/resource/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+pub mod default_user_secret;
+pub mod erlang_cookie;
+pub mod headless_service;
+pub mod rabbitmq_plugins;
+pub mod role;
+pub mod role_binding;
+pub mod service;
+pub mod service_account;
+pub mod stateful_set;
+
```
marshtompsxd, line 14: Perhaps create a prelude.rs and move the `pub use *` below to prelude.rs

src/controller_examples/rabbitmq_controller/exec/resource/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+pub mod default_user_secret;
+pub mod erlang_cookie;
+pub mod headless_service;
+pub mod rabbitmq_plugins;
+pub mod role;
+pub mod role_binding;
+pub mod service;
+pub mod service_account;
+pub mod stateful_set;
+
```
marshtompsxd, line 14: Perhaps create a prelude.rs and move the `pub use *` below to prelude.rs

src/controller_examples/rabbitmq_controller/exec/resource/default_user_secret.rs
```diff
@@ -0,0 +1,121 @@
...
+{
+    let mut data = StringMap::empty();
+    data.insert(new_strlit("username").to_string(), new_strlit("user").to_string());
+    data.insert(new_strlit("password").to_string(), new_strlit("changeme").to_string());
+    data.insert(new_strlit("type").to_string(), new_strlit("rabbitmq").to_string());
+    data.insert(new_strlit("host").to_string(),
+            rabbitmq.name().unwrap().concat(new_strlit(".")).concat(rabbitmq.namespace().unwrap().as_str()).concat(new_strlit(".svc"))
+    );
+    data.insert(new_strlit("provider").to_string(), new_strlit("rabbitmq").to_string());
+    // TODO: check \n
```
marshtompsxd, line 115: What is this TODO?

src/controller_examples/rabbitmq_controller/spec/resource/mod.rs
euclidgame: Do you learn this from verus or this is a better practice?

marshtompsxd: Learned from Verus. But I don't think it would affect much so it's ok to leave it as it is.

src/controller_examples/rabbitmq_controller/spec/resource/common.rs
```diff
@@ -0,0 +1,91 @@
...
+use crate::rabbitmq_controller::spec::types::*;
+use crate::reconciler::spec::{io::*, reconciler::*};
+use crate::state_machine::{action::*, state_machine::*};
+use crate::temporal_logic::defs::*;
+use vstd::prelude::*;
+use vstd::string::*;
+
+verus! {
+
+pub trait ResourceBuilder<T> {
```
marshtompsxd, line 21: Could you add some documentation to explain the interfaces here? And how do you expect users to use it (I assume this is designed for workflow like: Get -> Create/Update)?

src/controller_examples/rabbitmq_controller/spec/resource/mod.rs
euclidgame: Ok. I learn this from the k8s openapi :-))

src/controller_examples/rabbitmq_controller/spec/resource/common.rs
euclidgame: This is because rabbitmq only has such flow; if more patterns are included, the interface can also be extended.

euclidgame: And there are some things (about the design) to ponder

marshtompsxd:
I will merge it after the conflicts are resolved

src/controller_examples/rabbitmq_controller/exec/resource/default_user_secret.rs
euclidgame: Because I am not sure if there is a line break in the end of `"default_user = user\ndefault_pass = changeme"`

marshtompsxd:
@euclidgame Pls briefly describe how you refactor and deduplicate code in the description.
And feel free to merge it once you have no more changes.

PR #317, opened 2023-09-27
List all the invariants needed and organize them in a correct way

euclidgame:
This PR writes the spec for all the invariants needed to prove the liveness property of rabbitmq controller. There are some difference from the previous proof. The most point is that we don't update the entire spec for the sub resources in out implementation now, so we can't blindly describe the update request message.

Another thing is that I do something to deduplicate the code. First, I write reusable spec methods that can be used for all sub resources. Second, I add `created_or_updated_obj_matches_desired_state` lemma to the resource builder trait, so that we can use `ResourceBuilder::make` and `ResourceBuilder::update` methods to describe the create and update request message respectively (then `created_or_updated_obj_matches_desired_state` shows the resource reaches desired state).


marshtompsxd:
For the PR title, can you capitalize the first word only?

src/controller_examples/rabbitmq_controller/proof/helper_invariants/predicate.rs
```diff
@@ -0,0 +1,157 @@
...
+            &&& made_object.is_Ok()
+            &&& msg.content.get_create_request().obj == made_object.get_Ok_0()
+        }
+            // A reminder: The last predicate implies:
+            // && msg.content.get_create_request().obj.metadata.owner_references == Some(seq![rabbitmq.controller_owner_ref()])
+    }
+}
+
+/// We have to impose constraint on the resource version of update request message now instead of making no distinction
+/// to update request messages before. This is because we don't update the whole spec/data for the object now (we only update
```
marshtompsxd, line 83: > we don't update the whole spec/data for the object now

My understanding is that for the other fields (which are supposed to be constant, like selector of stateful set) we are going to prove a "safety" invariant saying that as long as the objects exist these fields equal X. Are you trying to do the same reasoning here (I didn't fully understand your comment here).

marshtompsxd:
@euclidgame You might need to rebase and solve some conflicts here.

PR #328, opened 2023-10-02
Add unit tests for pod/pod_spec/pod_temp_spec

ZichengMa:
In the PodSpec's tests, the Volume API is used. The unit tests for Volume will be in next PR. There are many different APIs should be tested for Volume, so I split those tests to next PR.


src/kubernetes_api_objects/volume.rs
```diff
@@ -29,6 +29,16 @@ impl Volume {
         }
     }
 
+    #[verifier(external_body)]
+    pub fn clone(&self) -> (volume: Volume)
+        ensures
+            volume@ == volume@,
```
marshtompsxd, line 35: postcondition is wrong

src/unit_tests/kubernetes_api_objects/mod.rs
```diff
@@ -13,3 +13,6 @@ pub mod probe;
 pub mod resource_requirements;
 pub mod tcp_socket_action;
 pub mod volume_mount;
+pub mod pod;
+pub mod pod_spec;
+pub mod pod_template_spec;
```
marshtompsxd, line 18: You can add a new empty line at the end of file

PR #340, opened 2023-10-09
Prove helper invariants for rabbitmq

euclidgame:


marshtompsxd:
Does the current proof consider the status updates (from sts and ds controller)?

src/kubernetes_cluster/proof/cluster_safety.rs
```diff
@@ -40,7 +40,10 @@ pub open spec fn etcd_object_is_well_formed(key: ObjectRef) -> StatePred<Self> {
             else if key.kind == RoleView::kind() { RoleView::unmarshal(s.resources()[key]).is_Ok() }
             else if key.kind == SecretView::kind() { SecretView::unmarshal(s.resources()[key]).is_Ok() }
             else if key.kind == ServiceView::kind() { ServiceView::unmarshal(s.resources()[key]).is_Ok() }
-            else if key.kind == StatefulSetView::kind() { StatefulSetView::unmarshal(s.resources()[key]).is_Ok() }
+            else if key.kind == StatefulSetView::kind() { 
+                &&& StatefulSetView::unmarshal(s.resources()[key]).is_Ok() 
+                &&& StatefulSetView::unmarshal(s.resources()[key]).get_Ok_0().state_validation()
```
marshtompsxd, line 45: It is a bit surprising that we need to say the state_validation holds for stateful set here, but do not say it holds for other types like role binding or role. Can you comment on why?

src/temporal_logic/rules.rs
```diff
@@ -2667,6 +2689,60 @@ macro_rules! leads_to_always_combine_n_with_equality_internal {
 pub use leads_to_always_combine_n_with_equality;
 pub use leads_to_always_combine_n_with_equality_internal;
 
+pub proof fn leads_to_always_tla_forall<T, A>(spec: TempPred<T>, p: TempPred<T>, a_to_p: FnSpec(A)->TempPred<T>, domain: Set<A>)
```
marshtompsxd, line 2692: Could you add some comments for this lemma on its common use case?

src/vstd_ext/map_lib.rs
```diff
@@ -25,4 +25,36 @@ pub proof fn union_prefer_right_self_changes_nothing<K, V>()
             map.union_prefer_right(map) =~= map,
 {}
 
+pub proof fn lemma_values_finite<K, V>(m: Map<K, V>)
```
marshtompsxd, line 28: Can you leave a TODO saying that we will use the Verus native lemma_values_finite once we update to the most recent Verus?

src/controller_examples/rabbitmq_controller/proof/safety/archive.rs
marshtompsxd, line 1: Seems that you deleted something from safety.rs. What are those?

src/kubernetes_cluster/proof/cluster_safety.rs
euclidgame: No longer needed and I will delete it

src/vstd_ext/map_lib.rs
euclidgame: Sure. This lemma was just added for my use

src/controller_examples/rabbitmq_controller/proof/helper_invariants/proof.rs
```diff
@@ -0,0 +1,1359 @@
...
+    proof::{predicate::*, resource::*},
+    spec::types::*,
+};
+use crate::temporal_logic::{defs::*, rules::*};
+use crate::vstd_ext::{multiset_lib, seq_lib, string_view::*};
+use vstd::{multiset::*, prelude::*, string::*};
+
+verus! {
+
+pub proof fn lemma_always_cr_objects_in_etcd_satisfy_state_validation(spec: TempPred<RMQCluster>)
```
marshtompsxd, line 25: Seems to me this can be made generic to all controllers?

src/controller_examples/rabbitmq_controller/proof/safety/archive.rs
euclidgame: I just needed some lemmas in it

src/controller_examples/rabbitmq_controller/proof/helper_invariants/proof.rs
```diff
@@ -0,0 +1,1359 @@
...
+        spec.entails(always(lift_action(RMQCluster::next()))),
+    ensures
+        spec.entails(always(lift_state(cr_objects_in_etcd_satisfy_state_validation()))),
+{
+    let inv = cr_objects_in_etcd_satisfy_state_validation();
+    RabbitmqClusterView::marshal_status_preserves_integrity();
+    init_invariant(spec, RMQCluster::init(), RMQCluster::next(), inv);
+}
+
+pub proof fn lemma_always_the_object_in_schedule_satisfies_state_validation(spec: TempPred<RMQCluster>)
```
marshtompsxd, line 37: Seems to me this can be made generic to all controllers?

src/controller_examples/rabbitmq_controller/proof/helper_invariants/proof.rs
euclidgame: No

src/controller_examples/rabbitmq_controller/proof/helper_invariants/proof.rs
```diff
@@ -0,0 +1,1359 @@
...
+    lemma_always_cr_objects_in_etcd_satisfy_state_validation(spec);
+    combine_spec_entails_always_n!(
+        spec, lift_action(stronger_next),
+        lift_action(RMQCluster::next()),
+        lift_state(cr_objects_in_etcd_satisfy_state_validation())
+    );
+    init_invariant(spec, RMQCluster::init(), stronger_next, inv);
+}
+
+pub proof fn lemma_always_the_object_in_reconcile_satisfies_state_validation(spec: TempPred<RMQCluster>)
```
marshtompsxd, line 58: Same as above

euclidgame: I just think we discussed this before...

euclidgame: And I don't want to add those ugly pre conditions so I wrote them for rabbitmq controller

euclidgame: Since these lemmas are pretty short, so it's also fine to copy them

src/controller_examples/rabbitmq_controller/proof/helper_invariants/proof.rs
marshtompsxd: Can you remind me on why?

marshtompsxd:
@euclidgame pls ping me again when you no longer have incoming commits

src/controller_examples/rabbitmq_controller/proof/liveness/proof.rs
```diff
@@ -28,6 +28,63 @@ use vstd::{prelude::*, string::*};
...
+{
+    assert forall |rabbitmq: RabbitmqClusterView| rabbitmq.well_formed()
+    implies #[trigger] cluster_spec().entails(liveness(rabbitmq)) by {
+        liveness_proof(rabbitmq);
+    };
+}
+
+proof fn liveness_proof(rabbitmq: RabbitmqClusterView)
+    requires
+        rabbitmq.well_formed(),
```
marshtompsxd, line 44: Actually, do we really need `rabbitmq.well_formed()` throughout the proof since the reconcile_core spec always assumes rabbitmq is well formed?

marshtompsxd: Of course I might miss something here

marshtompsxd: I am asking this coz I am thinking about removing the well formed thing from the top level liveness property, unless it will break the current proof.

src/controller_examples/rabbitmq_controller/proof/helper_invariants/unchangeable.rs
```diff
@@ -0,0 +1,265 @@
...
+        forall |msg: RMQMessage|
+            #[trigger] s.in_flight().contains(msg)
+            && resource_update_request_msg(resource_key)(msg)
+            && s.resources().contains_key(resource_key)
+            && s.resources()[resource_key].metadata.resource_version == msg.content.get_update_request().obj.metadata.resource_version
+            ==> unchangeable(sub_resource, msg.content.get_update_request().obj, rabbitmq)
+    }
+}
+
+proof fn lemma_always_object_in_every_create_request_msg_satisfies_unchangeable(
```
marshtompsxd, line 52: Can you also document this lemma and the one for update request? My understanding is that since the update request only updates part of the object, but our liveness property talks about the entire `.spec`, so you need some invariant to say the part, which never gets updated once created, remains unchanged. Is that correct?

src/controller_examples/rabbitmq_controller/proof/helper_invariants/validation.rs
```diff
@@ -0,0 +1,352 @@
...
+    spec::{reconciler::*, resource::*, types::*},
+};
+use crate::reconciler::spec::{reconciler::*, resource_builder::*};
+use crate::temporal_logic::{defs::*, rules::*};
+use crate::vstd_ext::{multiset_lib, seq_lib, string_view::*};
+use vstd::{multiset::*, prelude::*, string::*};
+
+verus! {
+
+pub open spec fn stateful_set_satisfies_unchangeable(obj: DynamicObjectView, rabbitmq: RabbitmqClusterView) -> bool {
```
marshtompsxd, line 31: Could you tell me why do we have lemmas specialized for stateful set here?

src/controller_examples/rabbitmq_controller/proof/helper_invariants/validation.rs
```diff
@@ -0,0 +1,352 @@
...
+    let key = rabbitmq.object_ref();
+    let sts_key = StatefulSetBuilder::get_request(rabbitmq).key;
+    |s: RMQCluster| {
+        s.resources().contains_key(key)
+        && s.resources().contains_key(sts_key)
+        ==> stateful_set_satisfies_unchangeable(s.resources()[sts_key], RabbitmqClusterView::unmarshal(s.resources()[key]).get_Ok_0())
+    }
+}
+
+pub proof fn lemma_always_stateful_set_in_etcd_satisfies_unchangeable(spec: TempPred<RMQCluster>, rabbitmq: RabbitmqClusterView)
```
marshtompsxd, line 53: Is this lemma used anywhere (I might miss something)?

src/controller_examples/rabbitmq_controller/spec/resource/stateful_set.rs
```diff
@@ -75,12 +75,6 @@ impl ResourceBuilder<RabbitmqClusterView, RabbitmqReconcileState> for StatefulSe
         let made_spec = make_stateful_set(rabbitmq, new_strlit("")@).spec.get_Some_0();
         &&& StatefulSetView::unmarshal(object).is_Ok()
         &&& sts.spec.is_Some()
-        &&& made_spec == StatefulSetSpecView {
```
marshtompsxd, line 78: Why is this deleted? I thought you needed it to say the fields other than replicas, templated and retention policy remains the same as the one in make_stateful_set

src/controller_examples/rabbitmq_controller/proof/helper_invariants/validation.rs
euclidgame: I am not separating stateful set--you can see `lemma_always_object_in_etcd_satisfies_unchangeable` applies to all subresources. Maybe the naming here is a bit confusing.

src/controller_examples/rabbitmq_controller/proof/helper_invariants/validation.rs
euclidgame: https://github.com/vmware-research/verifiable-controllers/blob/cc3a29a71b6a0bb49bf335de022eb0d824cd6331/src/controller_examples/rabbitmq_controller/proof/liveness/spec.rs#L434

marshtompsxd: Can you also document the lemmas here like the one in unchangeable.rs? From the name, I guess the lemma is for the validation rule but I am not sure. Also I am surprised by the removal of this https://github.com/vmware-research/verifiable-controllers/pull/340/files#r1354146855. Can you comment on it?

euclidgame: I use the `unchangeable` spec to describe what properties hold all through the executions. But the things that I deleted are not like that.

euclidgame: This lemma is to show that some fields of stateful set can be determined by that cr object its owner reference points to. It can be used to show the state of stateful set finally always matches the `rabbitmq` because certain fields are not updated by `update` method.

euclidgame: Since those fields stay unchanged, this invariant also helps to show the transition validation must be satisfied when proving `AtStep(Update, StatefulSet) ~> resource_match` (otherwise that update request will be rejected).

marshtompsxd: > I use the unchangeable spec to describe what properties hold all through the executions. But the things that I deleted are not like that so I removed that.

But the spec you deleted actually said that the fields (other than replicas, template and retention policy) are always the same as the ones from make_stateful_set through the executions. Note that I am not against your change, I am just trying to understand your rationale behind the change so I can better take over.

euclidgame: > But the spec you deleted actually said that the fields (other than replicas, template and retention policy) are always the same as the ones from make_stateful_set through the executions.

That doesn't hold all the time so I deleted them

marshtompsxd: > This lemma is to show that some fields of stateful set can be determined by that cr object its owner reference points to.

My confusion is mainly because I cannot really tell why we need such lemmas for the stateful set, but not for other sub-resources -- or did I miss something major here? Could you document the reason?

marshtompsxd: Here is my guess: there are three types of fields in terms of how they get (un)changed: (1) the fields whose values are fixed and immutable for whatever owner cr, (2) the fields whose values are decided by the owner cr but still immutable since created, (3) the fields whose values are mutable.

The unchangeable spec fn talks about (1), the lemmas in this file talk about (2) (and only stateful set has such fields), and (3) is handled when you apply wf1 (which is the easy case). Is that correct?

euclidgame: You are generally correct. I want to say we have the stateful set lemma is not only because of (2) but also because these fields won't be updated by the update methods.

euclidgame: I update the comments.

> (1) the fields whose values are fixed and immutable for whatever owner cr,
I also mentioned this in the comments of file `unchangeable`.

marshtompsxd: > also because these fields won't be updated by the update methods

Both the unchangeable lemma and the validation lemma here are designed to deal with this "partial update" behavior, which is fundamentally different from the early proof (where update always updates the entire object so wf1 solves all the problems regarding updates). This seems to be one of the major **new** challenges in the current proof, is my understanding correct?

euclidgame: Yes, one of. But I didn't make a summary to all the new changes.

src/controller_examples/rabbitmq_controller/spec/resource/stateful_set.rs
marshtompsxd: One last question on this: seems to me you can even make this spec fn return `true`, like the one for service object?

euclidgame: Why? I think leaving it here is fine

src/controller_examples/rabbitmq_controller/proof/helper_invariants/validation.rs
marshtompsxd: It would be very helpful if you can describe the major new changes to the proof (compared to the old first edition) in the description of this PR. I believe the reasoning on the cm's rv number is also one of them. You can just list the major new challenges at high level, and point to the lemma that solves the problem.

I also see that you made some progress on automating the previously tedious process of unpacking phased invariants. It would be great if you can also briefly document such in the description.

marshtompsxd: After that I will merge this PR (unless you have more to commit)

PR #341, opened 2023-10-11
Finish liveness proof for rabbitmq controller

marshtompsxd:
The core challenge is to find the right invariant to "combat" with the built-in stateful set controller that might update the stateful set('s status) at any point, thus incrementing the rv of the stateful set and failing the update request from the rabbitmq controller. The invariant introduced in #322 is not enough because it does not consider the case where the update request wants to establish dependencies between the stateful set object and other objects (like annotating the stateful set using a configmap's rv). This PR introduces a new invariant similar to the one introduced by #322 but also talks about rv of other objects used to update the stateful set. This invariant might be a little bit overfit for the config workflow in our controllers but it works well.

Besides the new invariant, we also adjust the proof structure a bit more to separate out the reasoning for the stateful set because it differs from other objects a lot: the stateful set is affected by the built-in controller, while others aren't (for now).

There are a list of to-dos for refactoring the proof. I leave these for the next PR(s):
1. Revise the unchangeable spec of certain objects (e.g., stateful set unchangeable can just be true)
2. Rename unchangeable spec
3. Add a new spec to talk about the fields that depend on the rabbitmq cr's spec but also immutable
4. Revise update logic for service object (should not update the cluster ip)
5. Revise update logic for role binding objects (should not update the role ref)
6. Strengthen "object in etcd is well formed" lemma to talk about state validation


src/kubernetes_cluster/proof/stateful_set_controller.rs
marshtompsxd, line 1: TODO: update the document, and remove the Mutable stuff from dynamic.rs

src/controller_examples/rabbitmq_controller/proof/liveness/resource_match.rs
marshtompsxd, line 1: Clean up and remove unused invariants

PR #347, opened 2023-10-15
Cosmetic changes to rabbitmq controller

marshtompsxd:
This PR makes the following changes:
1. Change the update logic for RoleBiding object to avoid updating its role_ref
2. Change the update logic for main and headless service objects to avoid updating its cluster ip
3. Fix the immutability bugs of storage_class_name and pod_management_policy: these two fields should not be nullable since they should be immutable (the immutable validation rules don't work well for nullable fields)


tianyin:
"Cosmetic changes"

PR #357, opened 2023-10-20
Improve performance

euclidgame:
Changes made by this PR:

- Change the triggers of some quantifiers. In this PR primarily switch the trigger `s.in_flight().contains(msg)` to something else.
- Remove redundant invariants or redundant predicates. Previously we have "each response only matches one pending request" and "only one response matches the pending request", but it should be better using "each pending request has unique id" and "each in flight message has unique id". The latter are more general and instinctive.
- Try to avoid using quantifiers. Using `forall` can give us stronger guarantees but it's needless in some cases. We don't need "every object in reconcile is well_formed or satisfies validation"; we just need "the object of key in reconcile is well_formed or satisfies validation".
- Use something like `assert_by` to hide some useless information. This PR doesn't include too much about this. But I found that using `assert_by` in `combine_spec_entails_always_n` (which is used nearly everywhere) really helps.


.github/workflows/ci.yml
marshtompsxd, line 1: Seems to be a mis-deletion?

.github/workflows/ci.yml
```diff
@@ -222,4 +222,4 @@ jobs:
       - name: Deploy zookeeper controller
         run: ./local-test.sh zookeeper
       - name: Run zookeeper e2e tests for ephemeral storage
-        run: cd e2e && cargo run -- zookeeper-ephemeral
+        run: cd e2e && cargo run -- zookeeper-ephemeral
```
marshtompsxd, line 225: Can you add the last empty line back?

marshtompsxd:
@euclidgame after fixing the zk verification failure, can you update me with the verification time before vs after the PR?

euclidgame:
The verification time is quite unstable

marshtompsxd:
> The verification time is quite unstable

How about the verification of running
`lemma_eventually_always_cm_rv_is_the_same_as_etcd_server_cm_if_cm_updated`
`lemma_eventually_always_every_resource_update_request_implies_at_after_update_resource_step`
`lemma_always_stateful_set_in_update_req_has_at_least_one_replica`
alone?

tianyin:
> The verification time is quite unstable

Can we run multiple times (say 10 or even 20) and then compare the distribution (or take the median)?

euclidgame:
> Can we run multiple times (say 10 or even 20) and then compare the distribution (or take the median)?

I can do this. @marshtompsxd , does this way make sense?

marshtompsxd:
> > Can we run multiple times (say 10 or even 20) and then compare the distribution (or take the median)?
> 
> I can do this. @marshtompsxd , does this way make sense?

yes

marshtompsxd:
@euclidgame Can you also update the description with how you improved the performance? Just like #352

euclidgame:
@marshtompsxd , added the description.

PR #380, opened 2023-10-23
Make some fields optional in FluentBitSpec

euclidgame:


marshtompsxd:
@euclidgame Can you confirm if these fields are mutable (in PodSpec) before merging this one?

PR #390, opened 2023-10-25
Add security_context to fluent controller

euclidgame:


src/controller_examples/fluent_controller/fluentbit/exec/resource/daemon_set.rs
```diff
@@ -306,6 +306,9 @@ fn make_fluentbit_pod_spec(fb: &FluentBit) -> (pod_spec: PodSpec)
     pod_spec.overwrite_dns_policy(fb.spec().dns_policy());
     pod_spec.overwrite_priority_class_name(fb.spec().priority_class_name());
     pod_spec.overwrite_scheduler_name(fb.spec().scheduler_name());
+    if fb.spec().security_context().is_some() {
```
marshtompsxd, line 309: Maybe we should make it consistent (**in another PR**) on whether to use `overwrite_xxx` or check `is_some` like this field. I prefer the latter (like this PR).

PR #410, opened 2023-11-11
Validate zk config for config map

euclidgame:
This PR strengthens the spec of `zk_api` by adding validation of config map. Specifically, we check the existence of config map in etcd and require that the value of "zoo.cfg" in its data is generated from `make_zk_config` with some valid `zk` as follows:
```rust
pub open spec fn validate_config_map_data(data: Map<StringView, StringView>) -> bool {
    let zk_config = data[new_strlit("zoo.cfg")@];
    &&& data.contains_key(new_strlit("zoo.cfg")@)
    &&& exists |zk: ZookeeperClusterView| 
        zk.state_validation()
        && (#[trigger] make_zk_config(zk)) == zk_config
}
``` 

We treat `make_zk_config` as trusted (details can be seen in option c of #415 ). We then fix the proof correspondingly.


src/controller_examples/zookeeper_controller/spec/zookeeper_api.rs
```diff
@@ -120,6 +122,39 @@ impl ExternalAPI for ZKAPI {
...
+    };
+    let cm_data = ConfigMapView::unmarshal(resources[cm_key]).get_Ok_0().data.get_Some_0();
+    &&& resources.contains_key(cm_key)
+    &&& ConfigMapView::unmarshal(resources[cm_key]).is_Ok()
+    &&& ConfigMapView::unmarshal(resources[cm_key]).get_Ok_0().data.is_Some()
+    &&& cm_data.contains_key(new_strlit("zoo.cfg")@)
+    &&& forall |min_i, min_j, max_i, max_j, min, max|
+            cm_data[new_strlit("zoo.cfg")@].subrange(min_i,min_j) == new_strlit("minSessionTimeout=")@ + int_to_string_view(min)
+            && cm_data[new_strlit("zoo.cfg")@].subrange(max_i,max_j) == new_strlit("maxSessionTimeout=")@ + int_to_string_view(max)
+            ==> min <= max
```
marshtompsxd, line 139: #405 also introduces a validation rule for non-zero timeout, can you also add that here?

euclidgame: Yes, added

marshtompsxd:
To prove liveness, you will also need to add the validation rules to the ZooKeeperCluster object (the spec version) I believe.

euclidgame:
@marshtompsxd , yes, but I think it belongs to the proof part, so I don't do it in this PR

marshtompsxd:
> @marshtompsxd , yes, but I think it belongs to the proof part, so I don't do it in this PR

Can we also do the proof in this PR? The current spec looks good to me. Just want to make sure the proof is never broken or external in main branch.

euclidgame:
Yes

euclidgame:
I seem to have changed the Cargo.lock at some point, I will revert it

src/controller_examples/zookeeper_controller/spec/zookeeper_api.rs
```diff
@@ -120,6 +122,50 @@ impl ExternalAPI for ZKAPI {
     }
 }
 
+pub open spec fn validate_config_map_data(data: Map<StringView, StringView>) -> bool {
```
marshtompsxd, line 125: Could you document this spec function?

src/vstd_ext/string_view.rs
```diff
@@ -24,6 +24,26 @@ pub proof fn int_to_string_view_injectivity()
         forall |i: int, j: int| int_to_string_view(i) == int_to_string_view(j) ==> i == j,
 {}
 
+#[verifier(external_body)]
+pub proof fn int_to_string_view_ensures()
+    ensures
+        forall |n: int| #![trigger int_to_string_view(n)]{
+            forall |i| #![trigger int_to_string_view(n)[i]]
+                0 <= i < int_to_string_view(n).len() ==> {
+                    &&& int_to_string_view(n)[i] == '0'
```
marshtompsxd, line 33: It should be `|||` rather than `&&&`

marshtompsxd:
@euclidgame could you update the description with how we improve the zk api spec with the checker on the configmap?

marshtompsxd:
After that I will merge it

euclidgame:
@marshtompsxd , updated.

PR #422, opened 2023-11-17
Count unit tests

ZichengMa:
1. Add `tools/count-tests.py`.
2. Use `python3 tools/count-tests.py` to show the coverage of unit-tests.
3. All the test files' names are now following to snake rule.
4. Add some missing tests.


marshtompsxd:
@ZichengMa can you explain a bit on how you count the test coverage?

tools/count-tests.py
```diff
@@ -0,0 +1,79 @@
...
+                structs = ['Error']
+            # print(f"\n{filename}: {structs}")
+            for struct_name in structs:
+                test_filename = f"{camel_to_snake(struct_name)}.rs"
+                test_file_path = os.path.join(test_directory_path, test_filename)
+            count_external_body, count_external_pub_fn, count_external_fn = count_functions(file_path)
+            total_count = count_external_body + count_external_pub_fn + count_external_fn
+            if not os.path.exists(test_file_path) and total_count > 0:
+                # test_count = count_test_functions(test_file_path)
+                print(f"\nThis file can not be found:{test_filename}\n")
```
marshtompsxd, line 74: I am a bit confused here. Seems that you are only using the last `test_filename` from the `for struct_name in structs:` loop?

ZichengMa:
For every rs file in the kubernetes_api directory, the script will firstly find all the structs' name and turn them from camel format into snake format and use this snake format to find the corresponding testing files in the unit_tests directory. For each test file, find the number of `pub fn test` but `test_kube` and `test_marshal` will be counted twice. since they are corresponding to `from_kube/into_kube` or `marshal/unmarshal`.
For api files in kubernetes_api directory, the script will count the number of `#[verifier(external)]\nfn/#[verifier(external)]\npub fn/#[verifier(external_body)]\npub fn/` until the View struct appears.
For every rs file in kubernetes_api directory, the script will compare the number of functions should be tested and the number of functions tested. If they are not equal, the script will print out the number of functions should be tested for each struct.

tools/count-tests.py
ZichengMa: You are right, this is not correct, I will fix it.

PR #435, opened 2023-11-28
Add `ports` field to `FluentBitSpec`

euclidgame:


marshtompsxd:
Could you rebase and push again? This PR has some redundancy with the previous one.

euclidgame:
Some proof statements might be redundant.

src/controller_examples/fluent_controller/fluentbit/exec/resource/service.rs
```diff
@@ -156,4 +164,55 @@ pub fn make_service(fb: &FluentBit) -> (service: Service)
...
+            service_ports@.map_values(|p: ServicePort| p@),
+            Seq::new(i as nat, |k: int| model_resource::make_service_port(ports[k]@))
+        );
+    }
+    while i < ports.len()
+        invariant
+            i <= ports@.len(),
+            service_ports@.map_values(|p: ServicePort| p@) == Seq::new(i as nat, |k: int| model_resource::make_service_port(ports[k]@)),
+        ensures
+            service_ports@.map_values(|p: ServicePort| p@) == Seq::new(ports@.len(), |k: int| model_resource::make_service_port(ports[k]@)),
```
marshtompsxd, line 183: I am a bit surprised that you don't need to write `decrease` here. Did I miss anything?

marshtompsxd: Seems that the tutorial does not even have that https://verus-lang.github.io/verus/guide/while.html

euclidgame: Not sure I follow. But I thought `decreases` only exists in spec functions. And there's no recursion.

marshtompsxd: I am referring to this https://dafny.org/latest/DafnyRef/DafnyRef#sec-loop-termination
But I guess Verus does not support that yet

euclidgame: In this case the `i` is decreasing? But in our case `i` is increasing.

marshtompsxd: `ports@.len() - i`

marshtompsxd: You can try to add `decrease` clause to this loop and see what Verus will say

PR #437, opened 2023-11-28
Add service selector to fluentbit spec

euclidgame:


marshtompsxd:
LGTM. Could you add a test to our FluentBit e2e test? basically set a selector in the CR and check whether the Service actually has the same selector

PR #487, opened 2024-05-31
Add a skeleton for verified StatefulSet controller

codyjrivera:
This pull request provides a skeleton object and a stub controller implementation for a verified `StatefulSet` controller `VStatefulSet`. The controller spec and status objects are defined in `src/deps_hack/src/lib.rs`; these definitions follow the documentation [here](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/stateful-set-v1/). The stub controller may be run locally using the command `./local-test.sh v_stateful_set`.


deploy.sh
```diff
@@ -24,10 +25,17 @@ if [ "$registry" != "remote" ] && [ "$registry" != "local" ]; then
...
 
-if cd deploy/$1 && kubectl apply -f crd.yaml && kubectl apply -f rbac.yaml && kubectl apply -f deploy_$registry.yaml; then
+if [ "$app" == "v_stateful_set" ]; then
+    app_ns="vstatefulset"
+else
+    app_ns=$app
+fi
+
+## use imperative management for CRDs since metadata for PodTemplateSpec is too long.
+if cd deploy/$1 && kubectl create -f crd.yaml && kubectl apply -f rbac.yaml && kubectl apply -f deploy_$registry.yaml; then
```
codyjrivera, line 35: This change is necessary since `VStatefulSetSpec` contains a field `template` of type `PodTemplateSpec`, whose metadata makes the resulting CRD file too big for the declarative API.

PR #497, opened 2024-06-07
Add liveness specification for VReplicaSet controller

codyjrivera:
Here I state a basic liveness theorem, based on Eventually Stable Reconciliation, for ReplicaSets. To check whether the cluster state matches the spec, I gather up all pods that match the label selector `selector` and are owned by the ReplicaSet into a Verus set. I then check whether the cardinality of that set is equal to the `replicas` field.

### DIFF
--- a/src/controller_examples/v_replica_set_controller/trusted/liveness_theorem.rs
+++ b/src/controller_examples/v_replica_set_controller/trusted/liveness_theorem.rs
@@ -35,6 +35,7 @@ pub open spec fn resource_state_matches(vrs: VReplicaSetView, resources: StoredS
 pub open spec fn owned_selector_match_is(vrs: VReplicaSetView, resources: StoredState, key: ObjectRef) -> bool {
     let obj = resources[key];
     &&& resources.contains_key(key)
+    &&& obj.kind == PodView::kind()
     &&& obj.metadata.owner_references_contains(vrs.controller_owner_ref())
     &&& vrs.spec.selector.matches(obj.metadata.labels)
 }
--- a/src/controller_examples/v_replica_set_controller/trusted/liveness_theorem.rs
+++ b/src/controller_examples/v_replica_set_controller/trusted/liveness_theorem.rs
@@ -1,0 +1,42 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: MIT
+#![allow(unused_imports)]
+use crate::v_replica_set_controller::v_replica_set::trusted::{spec_types::*, step::*};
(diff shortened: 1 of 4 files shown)


src/controller_examples/v_replica_set_controller/trusted/liveness_theorem.rs
```diff
@@ -0,0 +1,43 @@
...
+    &&& pods.finite()
+    &&& pods.len() == vrs.spec.replicas.unwrap_or(0)
+}
+
+pub open spec fn owned_selector_match_is(vrs: VReplicaSetView, resources: StoredState, key: ObjectRef) -> bool {
+    let obj = resources[key];
+    &&& resources.contains_key(key)
+    &&& obj.kind == PodView::kind()
+    &&& obj.metadata.owner_references_contains(vrs.controller_owner_ref())
+    &&& vrs.spec.selector.matches(obj.metadata.labels.unwrap_or(Map::empty()))
```
marshtompsxd, line 40: @codyjrivera could you add one more line here
```
&&& obj.metadata.deletion_timestamp.is_None()
```
Basically we only want to count the pods that are NOT in terminating state.
See the new commit I just added today: https://github.com/vmware-research/verifiable-controllers/commit/0f03d84ad24a2e156e71a9bc58d69f348cf9fa3e

src/controller_examples/v_replica_set_controller/trusted/liveness_theorem.rs
```diff
@@ -0,0 +1,43 @@
...
+
+pub open spec fn current_state_matches(vrs: VReplicaSetView) -> StatePred<VRSCluster> {
+    |s: VRSCluster| {
+        resource_state_matches(vrs, s.resources())
+    }
+}
+
+pub open spec fn resource_state_matches(vrs: VReplicaSetView, resources: StoredState) -> bool {
+    let pods: Set<ObjectRef> = Set::new(|k: ObjectRef| owned_selector_match_is(vrs, resources, k));
+    &&& pods.finite()
```
marshtompsxd, line 31: Just curious: do we need this line since we have the next line that talks about the length of the pods?

codyjrivera: The file https://github.com/verus-lang/verus/blob/main/source/vstd/set.rs says that `len` is only meaningful when the set is finite: infinite sets are allowed.

marshtompsxd: OK

PR #498, opened 2024-06-10
Add model reconciler and prove conformance for vrs controller

codyjrivera:
- Add a model reconciler for the `VReplicaSet` controller, modeled after the executable reconciler.
- Prove the implementation conforms to the model


marshtompsxd:
@codyjrivera could you work on verifying the exec reconcile conforms to the spec reconcile in the same PR? See this commented postcondition https://github.com/vmware-research/verifiable-controllers/blob/main/src/controller_examples/v_replica_set_controller/exec/reconciler.rs#L81

PR #509, opened 2024-08-08
Simplify compositional proof obligations

marshtompsxd:


tchajed (approved):
This looks like a great simplification to me! It already solves the problem of having a global ShapeShifter, and instead parameterizing the producer proof over an arbitrary (`any`) environment.

PR #510, opened 2024-08-09
Extend the abstract state machine as an array of controllers

marshtompsxd:
This PR makes two major changes to the abstract state machine used for demonstrating compositional proof:
(1) the abstract state machine is no longer hardcoded to two controllers, but takes a sequence of controllers
(2) the first proof obligation no longer constrains *what the cluster is*. Instead, its precondition constrains *what the cluster does* (the producer runs in the cluster and there is no interference).

### DIFF
--- a/src/soundness/compositionality/proof.rs
+++ b/src/soundness/compositionality/proof.rs
@@ -51,9 +51,8 @@ proof fn consumer_property_holds<S, I>(spec: TempPred<S>)
 // To prove the above theorem, there are three proof obligations.
 
 // Proof obligation 1:
-// Producer is correct when running with the shape shifter assuming no interference.
-// In fact, this theorem is all you need if you only care about the producer, not the
-// consumer.
+// Producer is correct when running in any cluster where there is no interference.
+// This theorem is all you need if you only care about the producer, not the consumer.
 #[verifier(external_body)]
 proof fn producer_property_holds_if_no_interference<S, I>(spec: TempPred<S>, cluster: Cluster<S, I>, p_index: int)
     requires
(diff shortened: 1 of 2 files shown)


marshtompsxd:
@tchajed I like the current way of specifying *what a cluster does* instead of *what a cluster is*. I have two questions here:
(1) It seems to me that `producer_property_holds_if_no_interference`'s precondition
```
forall |input, s, s_prime| #[trigger] producers::<S, I>()[p_index].next(input)(s, s_prime) ==> cluster.next()(s, s_prime)
```
is a bit redundant, because the producer's fairness condition already implies that the producer runs in the cluster. I cannot immediately think of where I would need this precondition to reason about the producer's correctness. Did I miss anything?
(2) Maybe we should also develop a consumer's version of `xxx_property_holds_if_no_interference`, which states that the consumer's property holds in any cluster where no interference (to the consumer and all its producers) happens. And the proof of the consumer's consumer will be built upon this theorem.

tchajed:
The key is the second point you made - the consumer should be verified in any cluster which has the consumer and its dependencies running, and with no other interference; right now your proof only applies to `consumer_and_producers` and couldn't be used by some downstream controller that wants `consumer` as a dependency. You'll need an assumption that `producers[p_index].next() ==> cluster.next()` in that case since by default you don't know anything about the cluster.

tchajed:
As far as the PR is concerned I think it's fine to merge as-is with a general theorem for the producer and a specific theorem for consumer correctness.

I think your next goal, whether in this PR or a subsequent one, would be to have symmetric theorems for producer and consumer (similar assumptions and guarantees, and both proofs quantifying over the cluster). Once you do that it's also a good idea to _also_ have a specific instantiation of the general consumer theorem for a specific cluster, to make sure that there is _some_ cluster that satisfies the assumptions - otherwise it's possible you've made some mistake in the definitions and the assumptions are impossible to achieve.

PR #511, opened 2024-08-13
Generalize the top-level theorem of the compositional proof for any cluster

marshtompsxd:
This PR introduces symmetric lemmas for the consumer and producers parameterized by `cluster` so the compositional proof is no longer hard coded to any particular cluster. This means that the consumer's correctness spec can now be used by any downstream controller in any cluster where the consumer and producers are running and there is no interference.

The generalization comes with some cost: the top-level theorem `consumer_and_producers_are_correct` now seems a bit more complex with extra preconditions to constrain the behavior of other controllers in the input `cluster`. The last precondition (the super long one) says that other controllers will not interfere with the consumer or producers. To identify the "other controllers," we need to introduce the id of the consumer and producers: the id is just the index in the `cluster.controllers`, and any controller not pointed by the provided ids is identified as one of the "other controllers."

Besides the top-level theorem, another major change is in the non-interference invariant. Previously the invariant says that "no one interferes with the consumer/producer." Such an invariant is too strong to prove if we don't pose many constraints on the cluster. This PR changes the invariant to "controller identified by `good_citizen_id` does not interfere with the consumer/producer."

With the new proof strategy, to prove the top-level theorem, there are five proof obligations:
1. prove each producer is correct assuming no one interferes with this producer
2. prove the consumer is correct assuming no one interferes with the consumer and each producer is correct
3. prove the consumer does not interfere with each producer
4. prove each producer does not interfere with another producer
5. prove each producer does not interfere with the consumer

To ensure the preconditions of the top-level theorem `consumer_and_producers_are_correct` are reasonable, this PR also applies `consumer_and_producers_are_correct` to a concrete cluster with the consumer and producers only to prove their correctness in such a cluster. See `consumer_and_producers_are_correct_in_a_concrete_cluster`.

### DIFF
--- a/src/soundness/compositionality/proof.rs
+++ b/src/soundness/compositionality/proof.rs
@@ -13,70 +13,158 @@ spec fn consumer_property<S>() -> TempPred<S>;
 // The top level property of the producer controller (e.g., ESR)
 spec fn producer_property<S>(p_index: int) -> TempPred<S>;
 
-// The inv saying that no one interferes with the producer's reconcile
-spec fn no_one_interferes_producer<S, I>(cluster: Cluster<S, I>, p_index: int) -> StatePred<S>;
+// The inv asserts that the controller indexed by good_citizen_id does not interfere with the consumer's reconcile
+// This invariant, if opened, should state something like:
+// forall |message| message is sent by the controller indexed by good_citizen_id ==> message does not modify object X,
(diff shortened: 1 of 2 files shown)


marshtompsxd:
I like the way that we no longer reason about two controllers' transitions in the current framework, but I also wish to simplify the preconditions of the top-level theorem.

PR #522, opened 2024-09-12
Add PodEvent state machine to model random pod API requests

codyjrivera:
I add a state machine to the cluster called `PodEvent` which issues random pod create, delete, and update requests on pods, while it is enabled. I also model a transition which disables `PodEvent`.

However, as implemented, not only are the liveness proofs of the controllers broken (that is to be expected), but also some of the proofs involving the builtin controllers in the compound state machine. For example, https://github.com/anvil-verifier/anvil/blob/3d0e043c1227c31213d9d07c050e1de32db8b387/src/kubernetes_cluster/proof/daemon_set_controller.rs#L298C10-L298C89 does not go through. Similar proofs are marked with a comment.

### DIFF
--- a/src/kubernetes_cluster/spec/pod_event/state_machine.rs
+++ b/src/kubernetes_cluster/spec/pod_event/state_machine.rs
@@ -1,0 +1,118 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: MIT
+#![allow(unused_imports)]
+use super::types::*;
+use crate::external_api::spec::*;
+use crate::kubernetes_api_objects::spec::{pod::*, resource::*};
+use crate::kubernetes_cluster::spec::{cluster::Cluster, message::*};
+use crate::reconciler::spec::reconciler::Reconciler;
+use crate::state_machine::action::*;
+use crate::state_machine::state_machine::*;
+use crate::temporal_logic::defs::*;
+use vstd::{multiset::*, prelude::*};
+
+verus! {
+
+impl <K: CustomResourceView, E: ExternalAPI, R: Reconciler<K, E>> Cluster<K, E, R> {
+
+pub open spec fn create_pod() -> PodEventAction<E::Input, E::Output> {
+    Action {
+        precondition: |input: PodEventActionInput, s: PodEventState| {
+            &&& s.enabled
+            &&& input.obj.metadata.name.is_Some()
+            &&& input.obj.metadata.namespace.is_Some()
+            &&& input.obj.kind == PodView::kind()
+            &&& PodView::unmarshal(input.obj).is_Ok()
+        },
+        transition: |input: PodEventActionInput, s: PodEventState| {
+            let create_req_msg = Message::pod_event_req_msg(Message::create_req_msg_content(
+                input.obj.metadata.namespace.get_Some_0(),
+                input.obj,
+                input.rest_id_allocator.allocate().1
+            ));
+
+            let s_prime = s;
+            let output = PodEventActionOutput {
+                send: Multiset::singleton(create_req_msg),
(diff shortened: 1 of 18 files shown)


src/kubernetes_cluster/spec/pod_event/types.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+pub enum Step {
+    CreatePod(DynamicObjectView),
+    UpdatePod(DynamicObjectView),
+    DeletePod(DynamicObjectView),
+}
+
+pub struct PodEventActionInput {
+    pub obj: DynamicObjectView,
```
marshtompsxd, line 24: This obj should be a `Pod` and (1) you don't need the precondition on marshalling and (2) when you send out the request you will need to marshal the Pod into DynamicObjectView

src/kubernetes_cluster/spec/pod_event/state_machine.rs
```diff
@@ -0,0 +1,118 @@
...
+            let output = PodEventActionOutput {
+                send: Multiset::singleton(delete_req_msg),
+                rest_id_allocator: input.rest_id_allocator.allocate().0,
+            };
+            (s_prime, output)
+        },
+    }
+}
+
+// NOTE: We don't need to model update_status, since its effect can be modeled
```
marshtompsxd, line 68: We need one for update_status: update won't change anything in the `status` field; only update_status does that.

src/kubernetes_cluster/spec/pod_event/types.rs
```diff
@@ -0,0 +1,37 @@
...
+use crate::kubernetes_api_objects::spec::{dynamic::*, resource::*};
+use crate::kubernetes_cluster::spec::message::*;
+use crate::state_machine::action::*;
+use crate::state_machine::state_machine::*;
+use crate::temporal_logic::defs::*;
+use vstd::{multiset::*, prelude::*};
+
+verus! {
+
+pub struct PodEventState {
```
marshtompsxd, line 13: I suggest we make `pod_event_enabled` a field in the cluster state of the top level state machine instead of here. And the PodEventState can be just a `()`.

src/kubernetes_cluster/spec/pod_event/types.rs
```diff
@@ -0,0 +1,36 @@
...
+use crate::kubernetes_api_objects::spec::{pod::*, resource::*};
+use crate::kubernetes_cluster::spec::message::*;
+use crate::state_machine::action::*;
+use crate::state_machine::state_machine::*;
+use crate::temporal_logic::defs::*;
+use vstd::{multiset::*, prelude::*};
+
+verus! {
+
+pub struct PodEventState {}
```
marshtompsxd, line 13: @codyjrivera you can just define `PodEventState` as `()`.

codyjrivera: This is consistent with how `ClientState` is defined. I will change both.

src/kubernetes_cluster/spec/pod_event/state_machine.rs
```diff
@@ -0,0 +1,130 @@
...
+use crate::temporal_logic::defs::*;
+use vstd::{multiset::*, prelude::*};
+
+verus! {
+
+impl <K: CustomResourceView, E: ExternalAPI, R: Reconciler<K, E>> Cluster<K, E, R> {
+
+pub open spec fn create_pod() -> PodEventAction<E::Input, E::Output> {
+    Action {
+        precondition: |input: PodEventActionInput, s: PodEventState| {
```
marshtompsxd, line 20: @codyjrivera we can remove the precondition of this and other actions (just make it `true`). We don't care whether the pod_event issues a valid pod or not as we don't need it to make progress.

PR #570, opened 2024-10-25
Document controllers' race on version number

marshtompsxd:


discussion/fairness/fairness-on-controller-race.md
```diff
@@ -0,0 +1,59 @@
+When verifying interactions between controllers, we find it hard to specify the fairness assumptions that allow controllers to make progress while there are races on the resource version number of the same object (stored in etcd).
+
+## Background
+Kubernetes controllers often employ a pattern that it reads an object from etcd (which uses version control for stored objects) and then writes the updated object back to etcd.
```
tchajed, line 4: "version control" is easily confused with something like git here, maybe call it version-based concurrency control. You can also mention the version numbers behave like logical timestamps.

discussion/fairness/fairness-on-controller-race.md
```diff
@@ -0,0 +1,59 @@
...
+}
+```
+
+Assume that initially `P` holds true for the object in etcd and only $C$'s `write` invalidates `P`.
+That is, $C$ will retry its `write` until it succeeds.
+
+The problem is how to prove that $C$ eventually successfully writes the object (for at least once) with some fairness assumption.
+Since $C$ races with every controller in $S$ on the version number, the fairness assumption needs to imply that $C$ eventually wins the race (for at least once).
+
+We find that even strong fairness is not sufficient to prove liveness here. To see why, consider the following execution:
```
tchajed, line 44: Let's be a little more specific here that by default, Anvil assumes fairness of the controller's action. You might need to add some detail that the pseudocode above turns into two state machine actions, and there's network delay in between as well.

I think you should also say that neither weak nor strong fairness solve the problem.

PR #571, opened 2024-10-31
WIP: Prove VReplicaSet WF1 Lemmas for Create Path

codyjrivera:
Modulo some auxiliary lemmas, this PR provides proofs for WF1 lemmas on the Create path of the ReplicaSet controller.

The pull request is WIP since I'm rebasing on top of some of the new changes.

### DIFF
--- a/src/v2/controllers/vreplicaset_controller/proof/helper_lemmas.rs
+++ b/src/v2/controllers/vreplicaset_controller/proof/helper_lemmas.rs
@@ -13,6 +13,7 @@ use crate::vreplicaset_controller::{
     trusted::{liveness_theorem::*, spec_types::*, step::*},
     proof::{helper_invariants, predicate::*},
 };
+use crate::vstd_ext::seq_lib::*;
 use vstd::prelude::*;
 
 verus! {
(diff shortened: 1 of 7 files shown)


src/vstd_ext/seq_lib.rs
```diff
@@ -104,13 +104,23 @@ pub proof fn seq_unequal_preserved_by_add_auto<A>(suffix: Seq<A>)
 }
 
 #[verifier(external_body)]
-pub proof fn seq_pred_false_on_all_elements_implies_empty_filter<A>(s: Seq<A>, pred: spec_fn(A) -> bool)
-    requires forall |e: A| #![auto] s.contains(e) ==> !pred(e),
-    ensures s.filter(pred).len() == 0;
+pub proof fn seq_pred_false_on_all_elements_is_equivalent_to_empty_filter<A>(s: Seq<A>, pred: spec_fn(A) -> bool)
+    ensures (forall |e: A| #![auto] s.contains(e) ==> !pred(e)) <==> s.filter(pred).len() == 0;
```
marshtompsxd, line 108: nit: let's set a trigger instead of using auto, which makes it even harder to predict the behavior of the solver

src/v2/controllers/vreplicaset_controller/proof/liveness/api_actions.rs
```diff
@@ -81,6 +81,75 @@ pub proof fn lemma_api_request_outside_create_or_delete_loop_maintains_matching_
         _ => {}
     };
 }
+
+// TODO: Prove this.
+#[verifier(external_body)]
+pub proof fn lemma_api_request_not_made_by_vrs_maintains_matching_pods(
```
marshtompsxd, line 87: This is an important lemma. I am surprised that it's not easy to prove given that the non-interference invariant is already very strong.

src/v2/controllers/vreplicaset_controller/proof/helper_lemmas.rs
```diff
@@ -68,4 +69,51 @@ pub proof fn vrs_non_interference_property_equivalent_to_lifted_vrs_non_interfer
     );
 }
 
+// TODO: Prove this lemma.
+// Annoying sequence reasoning.
+#[verifier(external_body)]
+pub proof fn lemma_filtered_pods_set_equals_matching_pods(
```
marshtompsxd, line 75: @codyjrivera could you write more descriptive comments for this lemma?

codyjrivera: Yes -- I can write comments.

src/v2/controllers/vreplicaset_controller/proof/liveness/api_actions.rs
codyjrivera: I have not tried to prove this yet --- I can try quickly this evening, falling back to comments if needed.

codyjrivera: Went through basically right away.

src/v2/controllers/vreplicaset_controller/proof/helper_lemmas.rs
marshtompsxd: Is this one already proved or not? Coz I still see the external_body annotation.

codyjrivera: It's not proven yet -- I had proved the first of the conjuncts but not the other two -- I've outlined roughly what we need to do in the body.

marshtompsxd: OK if feel free to merge this one when you are ready. I've done my review

PR #574, opened 2025-01-13
support for deletion timestamp added, tests failed

Catoverflow:
This PR implemented the model & executable part for 
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/replicaset/replica_set.go#L721-L723

Test is still broken, working on fixing that

### DIFF
--- a/src/v2/controllers/vreplicaset_controller/exec/reconciler.rs
+++ b/src/v2/controllers/vreplicaset_controller/exec/reconciler.rs
@@ -99,6 +99,13 @@ pub fn reconcile_core(v_replica_set: &VReplicaSet, resp_o: Option<Response<VoidE
     let namespace = v_replica_set.metadata().namespace().unwrap();
     match &state.reconcile_step {
         VReplicaSetReconcileStep::Init => {
+            if v_replica_set.metadata().has_deletion_timestamp() {
+                let state_prime = VReplicaSetReconcileState {
+                    reconcile_step: VReplicaSetReconcileStep::Done,
+                    ..state
+                };
+                return (state_prime, None);
+            }
             let req = KubeAPIRequest::ListRequest(KubeListRequest {
                 api_resource: Pod::api_resource(),
                 namespace: namespace,
@@ -119,13 +126,6 @@ pub fn reconcile_core(v_replica_set: &VReplicaSet, resp_o: Option<Response<VoidE
             let pods_or_none = objects_to_pods(objs);
             if pods_or_none.is_none() {
                 return (error_state(state), None);
-            }
-            if v_replica_set.metadata().has_deletion_timestamp() {
-                let state_prime = VReplicaSetReconcileState {
-                    reconcile_step: VReplicaSetReconcileStep::Done,
-                    ..state
-                };
-                return (state_prime, None);
             }
             let pods = pods_or_none.unwrap();
             let filtered_pods = filter_pods(pods, v_replica_set);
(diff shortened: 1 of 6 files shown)


src/v2/controllers/vreplicaset_controller/exec/reconciler.rs
```diff
@@ -120,6 +120,13 @@ pub fn reconcile_core(v_replica_set: &VReplicaSet, resp_o: Option<Response<VoidE
             if pods_or_none.is_none() {
                 return (error_state(state), None);
             }
+            if v_replica_set.metadata().has_deletion_timestamp() {
```
marshtompsxd, line 123: Can you move it to the `Init` step? If the VReplicaSet has deletion timestamp, then there is no need to list the pods whatsoever.

marshtompsxd:
So the conformance proof passes but the liveness proof breaks, right?

tianyin:
@Catoverflow could you write some brief descriptions when you open PRs?

src/v2/controllers/vreplicaset_controller/exec/reconciler.rs
Catoverflow: The problem is I cannot align well the code block in k8s controller with the step in anvil. Take the line of code I want to model as an example, I'm not sure if we filter out the entire `func (rsc *ReplicaSetController) syncReplicaSet(ctx context.Context, key string)` where should I put the corresponding timestamp checking logic in anvil's model.

Do you have any suggestion?

Catoverflow:
> So the conformance proof passes but the liveness proof breaks, right?

Yes, 2 failed tests are

```
error: assertion failed
   --> v2/controllers/vreplicaset_controller/proof/liveness/resource_match.rs:573:87
    |
573 |     assert forall |s, s_prime| pre(s) && #[trigger] stronger_next(s, s_prime) implies pre(s_prime) || post(s_prime) by {
    |                                                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ assertion failed

note: function body check: not all errors may have been reported; rerun with a higher value for --multiple-errors to find other potential errors in this function
   --> v2/controllers/vreplicaset_controller/proof/liveness/resource_match.rs:460:1
    |
460 | / pub proof fn lemma_from_after_receive_list_pods_resp_to_send_create_pod_req(
461 | |     vrs: VReplicaSetView, spec: TempPred<ClusterState>, cluster: Cluster, controller_id: int,
462 | |     resp_msg: Message, diff: int
463 | | )
    | |_^

error: assertion failed
    --> v2/controllers/vreplicaset_controller/proof/liveness/resource_match.rs:1071:87
     |
1071 |     assert forall |s, s_prime| pre(s) && #[trigger] stronger_next(s, s_prime) implies pre(s_prime) || post(s_prime) by {
     |                                                                               
```

Catoverflow:
> @Catoverflow could you write some brief descriptions when you open PRs?

added

marshtompsxd:
@Catoverflow You might need to revise the premise of the ESR property to focus on reasoning about VReplicaSet that has no deletion timestamp. You can work with @codyjrivera on that.

src/v2/controllers/vreplicaset_controller/proof/helper_invariants/predicate.rs
```diff
@@ -328,10 +328,18 @@ pub open spec fn every_delete_request_from_vrs_has_rv_precondition_that_is_less_
         }
     }
 }
+
+pub open spec fn vrs_does_not_have_deletion_timestamp (
```
codyjrivera, line 332: You should separate the two parts of this predicate. The first line (talking about `s.resources`) and the second line (talking about `s.ongoing_reconciles`) should be in separate predicates, since we can assume the first holds in the definition of ESR, but we need to prove that the second holds.

PR #575, opened 2025-01-16
Prove ReplicaSet invariants, Convert Model from `usize` to unbounded Nats

codyjrivera:
Prove several additional invariants for VReplicaSet; convert reconcile step data structure to unbounded integers.


src/v2/controllers/vreplicaset_controller/model/reconciler.rs
```diff
@@ -1,14 +1,14 @@
...
-use vstd::prelude::*;
+use vstd::{prelude::*, math::abs};
 
 verus! {
 
 pub struct VReplicaSetReconciler {}
 
 pub struct VReplicaSetReconcileState {
-    pub reconcile_step: VReplicaSetReconcileStep,
+    pub reconcile_step: VReplicaSetRecStepView,
```
codyjrivera, line 11: In my judgement, `VReplicaSetReconcileStepView` is far too long of a name, so I named it `VReplicaSetRecStepView` instead. If you have a better suggestion, go ahead --- it's a matter of search and replace.

src/v2/controllers/vreplicaset_controller/model/reconciler.rs
```diff
@@ -97,7 +97,7 @@ pub open spec fn reconcile_core(v_replica_set: VReplicaSetView, resp_o: Option<R
                                 obj: pod.marshal(),
                             });
                             let state_prime = VReplicaSetReconcileState {
-                                reconcile_step: VReplicaSetReconcileStep::AfterCreatePod((diff - 1) as usize),
+                                reconcile_step: VReplicaSetRecStepView::AfterCreatePod(abs(diff - 1)),
```
marshtompsxd, line 100: Why do we need `abs` here? Is it to make sure the result is still a `nat` since there is subtraction here?

src/v2/controllers/vreplicaset_controller/proof/helper_invariants/predicate.rs
```diff
@@ -157,30 +108,43 @@ pub open spec fn no_pending_create_or_delete_request_not_from_controller_on_pods
...
-//
-// TODO: Prove this.
-//
-// Proving this for the VReplicaSet controller should be easy; we'd need to do a similar
-// proof for other state machines within the compound state machine.
-//
 
 pub open spec fn every_create_matching_pod_request_implies_at_after_create_pod_step(
-    vrs: VReplicaSetView, controller_id: int,
+    vrs: VReplicaSetView, cluster: Cluster, controller_id: int,
```
marshtompsxd, line 113: It seems that the only thing you need from `cluster` is `installed_types`? If so I suggest only pass `installed_types` to this function, instead of the entire `cluster`.

src/v2/kubernetes_cluster/proof/controller_runtime_safety.rs
```diff
@@ -90,6 +90,8 @@ pub open spec fn state_comes_with_a_pending_request(self, controller_id: int, st
     &&& forall |cr, resp_o, pre_state| #[trigger] state((self.controller_models[controller_id].reconcile_model.transition)(cr, resp_o, pre_state).0) ==> (self.controller_models[controller_id].reconcile_model.transition)(cr, resp_o, pre_state).1.is_Some()
 }
 
+// TODO: Proof exceeds default rlimit now, investigate.
+#[verifier(rlimit(10000))]
```
marshtompsxd, line 94: wow do we really need rlimit to be so large?

src/v2/controllers/vreplicaset_controller/proof/liveness/resource_match.rs
```diff
@@ -598,6 +605,7 @@ pub proof fn lemma_from_after_receive_list_pods_resp_to_send_create_pod_req(
     );
 }
 
+#[verifier(rlimit(10000))]
```
marshtompsxd, line 608: Maybe also add a todo here. Does this one really need so large rlimit?

src/v2/controllers/vreplicaset_controller/model/reconciler.rs
codyjrivera: The compiler complains if the expression is left as-is without a cast/operation. If you'd prefer `as nat`, I can do that change.

src/v2/controllers/vreplicaset_controller/proof/liveness/resource_match.rs
codyjrivera: They seem not to require extra `rlimit` now. Nevertheless, I'm marking them as flaky proofs.

src/v2/controllers/vreplicaset_controller/model/reconciler.rs
```diff
@@ -1,14 +1,14 @@
 use crate::kubernetes_api_objects::spec::prelude::*;
 use crate::reconciler::spec::{io::*, reconciler::*};
 use crate::vreplicaset_controller::trusted::{spec_types::*, step::*};
-use vstd::prelude::*;
+use vstd::{prelude::*, math::abs};
```
marshtompsxd, line 4: @codyjrivera Can you also remove this import of `abs`? After that you can merge it.

PR #577, opened 2025-01-16
Fix version for home

codyjrivera:
Fix `home` dependency (set to version "0.5.9") to address issue https://github.com/anvil-verifier/anvil/issues/576.


marshtompsxd:
@codyjrivera Can you also pin the home version in https://github.com/anvil-verifier/anvil/blob/main/src/deps_hack/Cargo.toml and see if it works?

marshtompsxd:
If that also does not work, try `cargo update home --precise 0.5.9` in the CI workflow before running tests.

marshtompsxd:
OK the final workaround works! @codyjrivera could you remove the pin in the two Cargo.toml then?

PR #585, opened 2025-02-01
Proved `seq_map_value_lemma`

Catoverflow:
- proved seq_map_value_lemma
- improved code style in `src/vstd_ext/seq_lib` and `src/vstd_ext/map_lib`


src/vstd_ext/seq_lib.rs
```diff
@@ -203,22 +208,15 @@ pub proof fn seq_filter_is_a_subset_of_original_seq<A>(s: Seq<A>, pred: spec_fn(
...
         let subseq = s.drop_last();
-        seq_filter_is_a_subset_of_original_seq(s.drop_last(), pred);
+        seq_filter_is_a_subset_of_original_seq(subseq, pred);
         assert(forall |i: int| 0 <= i < subseq.filter(pred).len() ==> subseq.contains(#[trigger] subseq.filter(pred)[i]));
         // assert(forall |i: int| 0 <= i < s.filter(pred).len() ==> s.contains(#[trigger] s.filter(pred)[i]));
         // assert(forall |e: A| s.filter(pred).contains(e) ==> #[trigger] s.contains(e));
     }
 }
 
-#[verifier(external_body)]
```
marshtompsxd, line 213: @Catoverflow Can you put `#[verifier(external_body)]` back? So it reminds us that this one has not been proved yet.

src/vstd_ext/seq_lib.rs
```diff
@@ -108,12 +108,17 @@ pub proof fn seq_pred_false_on_all_elements_is_equivalent_to_empty_filter<A>(s:
 {
     if s.len() != 0 {
         assert((forall |e: A| s.contains(e) ==> !pred(e)) ==> s.filter(pred).len() == 0) by {
-            assume(forall |e: A| s.contains(e) ==> !pred(e));
```
marshtompsxd, line 111: I didn't realize that we had `assume` in the proof. Is there any other `assume` used in the proof?

codyjrivera: I think you can replace this with 
```
if (forall |e: A| s.contains(e) ==> !pred(e)) {
    // proof
}
```
to the same effect. You shouldn't need to prove anything in the else case, since an implication $A \Rightarrow B$ is vacuously true whenever $A$ is false.

src/vstd_ext/seq_lib.rs
Catoverflow: @marshtompsxd This one automatically pass with no manual proof needed

src/vstd_ext/seq_lib.rs
Catoverflow: I think I accidentally introduced the `assume` back during merge, let me fix it

src/vstd_ext/seq_lib.rs
Catoverflow: @marshtompsxd It's added, I think all are resolved now

PR #587, opened 2025-02-05
Prove `lemma_filtered_pods_set_equals_matching_pods`

Catoverflow:
`lemma_filtered_pods_set_equals_matching_pods` is proved with 2 seq_lib lemma introduced:
- `pred_on_element_equal_to_pred_on_index`
- `seq_map_filter_equal_seq_filter_map`

To be proved later

PR #592, opened 2025-03-04
Prove vstd lemma of `push_to_set_eq_to_set_insert`, `map_values_to_set_eq_to_set_mk_map_values`, `lemma_mk_map_insert_k`, `lemma_k_v`

Catoverflow:
Proved following vstd lemma introduced in #587
- <s>`lemma_k_v`</s>
- `push_to_set_eq_to_set_insert`
- `map_values_to_set_eq_to_set_mk_map_values`
- `lemma_mk_map_insert_k`

### DIFF
--- a/src/vstd_ext/map_lib.rs
+++ b/src/vstd_ext/map_lib.rs
@@ -52,7 +52,4 @@ pub proof fn injective_finite_map_implies_dom_len_is_equal_to_values_len<K, V>(m
     }
 }
 
-pub broadcast proof fn lemma_k_v<K, V>(m: Map<K, V>, k: K, v: V)
-    ensures (#[trigger] m.contains_key(k) && #[trigger] m[k] == v) <==> #[trigger] m.contains_pair(k, v);
-
 }
--- a/src/vstd_ext/map_lib.rs
+++ b/src/vstd_ext/map_lib.rs
@@ -52,4 +52,7 @@ pub proof fn injective_finite_map_implies_dom_len_is_equal_to_values_len<K, V>(m
     }
 }
 
+pub broadcast proof fn lemma_k_v<K, V>(m: Map<K, V>, k: K, v: V)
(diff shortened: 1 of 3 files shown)


Catoverflow:
Seems CI workflow needs to be updated

src/vstd_ext/map_lib.rs
```diff
@@ -52,4 +52,7 @@ pub proof fn injective_finite_map_implies_dom_len_is_equal_to_values_len<K, V>(m
     }
 }
 
+pub broadcast proof fn lemma_k_v<K, V>(m: Map<K, V>, k: K, v: V)
```
marshtompsxd, line 55: @Catoverflow  I suggest we don't make this broadcast. Actually, do we really need this one? The definition of `contains_pair` is open and is exactly `self.dom().contains(k) && self[k] == v`

Catoverflow: @marshtompsxd Fixed and passed

tianyin: I think @Catoverflow is having fun with this...

marshtompsxd:
@Catoverflow In this case, you can rebase `main` branch and (force) push again, so that your PR won't contain the content of the previous PR that fixes the CI

Catoverflow:
@marshtompsxd I think `rabbitmq-e2e-test` is flaky. We got a timeout error here:
https://github.com/anvil-verifier/anvil/actions/runs/13727089105/job/38395923455

In #594 the related change is I updated OS/rust toolchain version, which may be an indirect cause of that

marshtompsxd:
Debugging flaky test is always hard, because they are flaky

Catoverflow:
> Debugging flaky test is always hard, because they are flaky

@marshtompsxd Is it possible to rewrite this piece of test to prevent timeout at least? If the update only makes it slower.

PR #594, opened 2025-03-05
Fix broken CI, add cache support for Verus build result, update CI OS

Catoverflow:
- Fixed broken CI caused by rust toolchain update
- Added cache support for Verus build
  - As we checked out the commit ID the cache will always be reused, so we could save 3mins for every jobs in CI
- Bumped Github action API call version and runner OS version
  - Cache for Go environment is enabled by default since V4

However, as action cache does not support caching on relative path (we put Verus in `../verus` before), and there is no convenient way to get runner root dir instead of project root dir. There are 2 ways to workaround:
1. Export runner root dir in action runtime to action env, cache `${{env.runer_root}}/verus`, and reuse in each cache fetch
2. Just put verus in project root, which expands to `anvil/verus`. Require some runtime modification on CI jobs, together with runtime modification in `Cargo.toml`.

Currently I'm using the 2nd way.

BTW, if we switch to vargo from cargo we can(?) also skip the installation of rust toolchains for most of the jobs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,8 @@ jobs:
           ./tools/get-z3.sh
           . "$HOME/.cargo/env"
           . ../tools/activate
+          echo $VARGO_TOOLCHAIN
+          rustup show active-toolchain
           vargo clean
           vargo build --release
       - name: Cache verus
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -12,253 +12,186 @@ on:
...
+      - name: Install Rust toolchain
+        run: |
+          curl --proto '=https' --tlsv1.2 --retry 10 --retry-connrefused -fsSL "https://sh.rustup.rs" | sh -s -- --default-toolchain 1.82.0-x86_64-unknown-linux-gnu -y
+      - name: Restore build cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            verus
+          key: verus-${{ hashFiles('source/Cargo.lock') }}
+      - name: Verify fluent controllerc
```
marshtompsxd, line 53: nit: controllerc -> controller

marshtompsxd:
@Catoverflow Great job!

> BTW, if we switch to vargo from cargo we can(?) also skip the installation of rust toolchains for most of the jobs.

We can try it in the next PR

PR #595, opened 2025-03-12
Add proofs for delete-related invariants

codyjrivera:
Here I add proofs for `each_vrs_in_reconcile_implies_filtered_pods_owned_by_vrs` and `at_after_delete_pod_step_implies_filtered_pods_in_matching_pod_entries`.

~~The latter requires a new variable to be associated with the API server: `used_names`, which keeps track of every name which has been introduced there. This is accounted for with an expanded `generated_name_is_unique` guarantee, and an invariant saying everything in etcd has a name in `used_names`.~~

~~Removing this variable seems to greatly slow the proof down, but I cannot explicitly track why it's helpful.~~

`used_names` can be removed.


src/v2/kubernetes_cluster/spec/api_server/types.rs
```diff
@@ -11,6 +11,7 @@ pub struct APIServerState {
     pub resources: StoredState,
     pub uid_counter: Uid,
     pub resource_version_counter: ResourceVersion,
+    pub used_names: Set<StringView>,
```
marshtompsxd, line 14: @codyjrivera could you add a comment here to explain and remind ourselves in the future why we want this history variable?

src/v2/kubernetes_cluster/spec/api_server/state_machine.rs
```diff
@@ -251,7 +251,10 @@ pub closed spec fn generate_name(s: APIServerState) -> StringView;
 // For more details, see the implementation: https://github.com/kubernetes/kubernetes/blob/v1.30.0/staging/src/k8s.io/apiserver/pkg/registry/generic/registry/store.go#L432-L443
 #[verifier(external_body)]
 pub proof fn generated_name_is_unique(s: APIServerState)
-    ensures forall |key| #[trigger] s.resources.contains_key(key) ==> key.name != generate_name(s)
+    ensures 
+        forall |key| #[trigger] s.resources.contains_key(key) ==> key.name != generate_name(s),
+        forall |key: ObjectRef| #[trigger] s.used_names.contains(key.name) ==> key.name != generate_name(s),
```
marshtompsxd, line 256: @codyjrivera I am concerned about assuming `forall |key: ObjectRef| #[trigger] s.used_names.contains(key.name) ==> key.name != generate_name(s)`. Could you elaborate on why we need this assumption?

src/v2/kubernetes_cluster/spec/api_server/types.rs
marshtompsxd: btw, don't you need to initialize `used_names` to empty in the initial state?

codyjrivera: Stale, I removed `used_names`.

src/v2/controllers/vreplicaset_controller/proof/helper_invariants/predicate.rs
```diff
@@ -216,71 +220,110 @@ pub open spec fn each_vrs_in_reconcile_implies_filtered_pods_owned_by_vrs(contro
...
+                let req_msg = s.ongoing_reconciles(controller_id)[key].pending_req_msg.get_Some_0();
+                &&& s.ongoing_reconciles(controller_id).contains_key(vrs.object_ref())
+                &&& VReplicaSetReconcileState::unmarshal(s.ongoing_reconciles(controller_id)[vrs.object_ref()].local_state).is_ok()
+                &&& state.filtered_pods.is_Some()
+                &&& diff <= filtered_pod_keys.len()
+                &&& filtered_pod_keys.no_duplicates()
+                &&& s.ongoing_reconciles(controller_id)[triggering_cr.object_ref()].pending_req_msg.is_Some()
+                &&& req_msg.dst.is_APIServer()
+                &&& req_msg.content.is_delete_request()
+                &&& forall |i| #![auto] 0 <= i < diff ==> {
```
marshtompsxd, line 290: Can we avoid `auto` here?

src/v2/controllers/vreplicaset_controller/proof/helper_invariants/predicate.rs
```diff
@@ -216,71 +220,110 @@ pub open spec fn each_vrs_in_reconcile_implies_filtered_pods_owned_by_vrs(contro
...
             }
     }
 }
-//
-// TODO: Prove this.
-//
-// No hints.
-//
 
 pub open spec fn at_after_delete_pod_step_implies_filtered_pods_in_matching_pod_entries(
```
marshtompsxd, line 262: @codyjrivera is it possible to split this invariant into two: one talks about `AfterDeletePod` and one talks about `AfterListPods`? The current one is a mix of two.

src/v2/kubernetes_cluster/proof/network.rs
```diff
@@ -251,6 +251,64 @@ pub proof fn lemma_always_every_in_flight_req_msg_has_different_id_from_pending_
     init_invariant::<ClusterState>(spec, self.init(), stronger_next, invariant);
 }
 
+pub open spec fn every_in_flight_req_msg_has_different_id_from_all_pending_req_msg(controller_id: int) -> StatePred<ClusterState> {
```
marshtompsxd, line 254: Can we follow the naming of `every_in_flight_req_msg_has_different_id_from_pending_req_msg_of` and name this one as `every_in_flight_req_msg_has_different_id_from_pending_req_msg_of_every_ongoing_reconcile`?

src/v2/controllers/vreplicaset_controller/proof/helper_invariants/predicate.rs
codyjrivera: Maybe, but the severance is almost certainly outside the scope of what can be done in a single PR. Perhaps a comment on both of these invariants saying that they should be cleaned up?

marshtompsxd: Yes. Leave a todo here.

PR #598, opened 2025-03-28
Split admission controller code for verification

agrama64:
Split admission control implementation into three files main.rs, handlers.rs, validator.rs

validator.rs contains our admission logic split into smaller functions, which is what we seek to verify

(We also have included a commit that adds support for the CRD used in anvil)


src/controller_examples/v_replica_set_controller/admission_control/src/validator.rs
marshtompsxd, line 1: @agrama64 For the next step, you will need to:
1. write an executable version (in https://github.com/anvil-verifier/anvil/blob/main/src/v2/controllers/vreplicaset_controller/trusted/exec_types.rs) of the [state validation logic](https://github.com/anvil-verifier/anvil/blob/main/src/v2/controllers/vreplicaset_controller/trusted/spec_types.rs#L115). You can leave it unproved for now.
2. call the executable state validation logic in this file. The difference is that you will need to convert the json object into VReplicaSet defined in https://github.com/anvil-verifier/anvil/blob/main/src/v2/controllers/vreplicaset_controller/trusted/exec_types.rs. To do so, you will need to import the Verus files and need to use https://github.com/anvil-verifier/anvil/blob/main/src/v2/controllers/vreplicaset_controller/trusted/exec_types.rs#L82.
3. Finish the proof in bullet 1.

marshtompsxd: > you will need to import the Verus files

I recommend that you make the current admission controller part of the Anvil project, instead of a standalone Rust project. In other words, you will have an unverified entry point like https://github.com/anvil-verifier/anvil/blob/main/src/v2_vreplicaset_controller.rs; the entry point will call the core validation logic which is verified against the spec version [state validation logic](https://github.com/anvil-verifier/anvil/blob/main/src/v2/controllers/vreplicaset_controller/trusted/spec_types.rs#L115). And you will build the admission controller using the https://github.com/anvil-verifier/anvil/blob/main/build.sh script.

PR #600, opened 2025-04-06
Refactor the VRS admission controller to use verified admission logic

agrama64:
In this PR we implement an executable function called state_validation in src/v2/controllers/vreplicaset_controller/trusted/exec_types.rs.

This function is then invoked in src/v2/controllers/vreplicaset_controller/exec/validator.rs

We also have implemented two entry points: one for the original controller_examples directory where our work was previously located and one for v2 where we are migrating to.

We kept the old entrypoint just to try out some toy techniques for validating the number of replicas which is found in src/controller_examples/v_replica_set_controller/exec/validator.rs

There isn't much "proof" here per se outside of verify_replicas but it seems like everything runs without verus complaining


marshtompsxd:
@agrama64 Does the CI build and test the new admission controller https://github.com/anvil-verifier/anvil/pull/600/files#diff-42f04e907930cca56de7af4e62041bde9e208e6df6d22a766ef38020d9681f35? If not we should update the CI workflow

marshtompsxd:
@agrama64 Does the admission controller e2e test run the new implementation or the old one? We should test the new one.

fredred375:
@marshtompsxd 
Updated ci script to run e2e on new implementation instead

PR #601, opened 2025-04-09
Complete VRS liveness proof

Catoverflow:
@codyjrivera and I
Proved
- `vrs_in_ongoing_reconciles_does_not_have_deletion_timestamp`
- `lemma_true_leads_to_pod_monkey_always_disabled`
- `tla_forall_schedule_controller_reconcile_weak_fairness_is_stable`
- `tla_forall_external_next_weak_fairness_is_stable`

Completed
- `src/v2/controllers/vreplicaset_controller/proof/liveness/spec.rs`
- `src/v2/controllers/vreplicaset_controller/proof/liveness/proof.rs`
- `src/v2/kubernetes_cluster/proof/controller_runtime_safety.rs`

Fixed
- flaky and slow proof for `lemma_eventually_always_at_after_delete_pod_step_implies_filtered_pods_in_matching_pod_entries`

Simplified
- `reconcile_eventually_terminates`

Masked
- `lemma_from_pending_req_in_flight_at_some_state_to_in_flight_resp_matches_pending_req_at_some_state`
This require another PR <s>to update the cluster state machine</s>(there's a better way to do this by weakening the precondition to be the vrs controller will not send message to its non-existing external part). We have created a branch for it at `external-message-fix`

Now we are done (almost, after the masked one is fixed) for VReplicaSet controller.

PR #602, opened 2025-04-10
Handle external_model properly

codyjrivera:
I redefine `there_is_no_request_msg_to_external` to `there_is_no_request_msg_to_external_from_controller`, so we can (successfully) prove some as-yet unproven lemmas. In summary, we only need to ensure _our controller_ doesn't send external messages. The rename is justified because both the original name and definition put us on an unproductive tangent.

This marks the _actual_ completion of the proof for `VReplicaSet`, and ~~depends on Pull Request #601~~ which is already merged .


marshtompsxd:
@codyjrivera Fantastic!

PR #604, opened 2025-04-16
Add new TLA rule `transform_leads_to_with_until` for compositionality

marshtompsxd:


Catoverflow:
@marshtompsxd Would it be better to add a `until` defined as
```
pub open spec until<T>(self, b: TempPred<T>, next: TempPred<T>) ->
    always(self.and(next).implies(later(self).or(later(b))))
```

Catoverflow (approved):

PR #606, opened 2025-04-17
Add e2e tests for V2 VReplicaSet

codyjrivera:
These tests duplicate the functionality of the V1 tests as of now. Note that I have also adapted some scripts to convert between `snake_case`, as used in filenames, and `kebab-case`, as used in the .yaml files (see `v2_vr..` versus `v2-vr..`).


marshtompsxd:
"Beware of bugs in the above code; I have only proved it correct, not tried it."

PR #607, opened 2025-04-18
Specification of Rely-Condition for VReplicaSet

codyjrivera:
Here I specify the rely condition for VReplicaSet: no controller creates/updates/deletes pods that are owned by a VReplicaSet.

I don't yet know how many proofs this breaks: once I have an idea (by running and marking the appropriate proofs), I will document this in the PR.

### DIFF
--- a/src/v2/controllers/vreplicaset_controller/trusted/liveness_theorem.rs
+++ b/src/v2/controllers/vreplicaset_controller/trusted/liveness_theorem.rs
@@ -45,7 +45,7 @@ pub open spec fn vrs_not_interfered_by(other_id: int) -> StatePred<ClusterState>
                 let owner_references = req.obj.metadata.owner_references.get_Some_0();
                 &&& req.obj.kind == Kind::PodKind
                 &&& req.obj.metadata.owner_references.is_Some()
-                &&& exists |vrs: VReplicaSet| 
+                &&& exists |vrs: VReplicaSetView| 
                     #[trigger] owner_references.contains(vrs.controller_owner_ref())
             },
             // Other controllers don't try to update pods owned by a VReplicaSet.
@@ -57,7 +57,7 @@ pub open spec fn vrs_not_interfered_by(other_id: int) -> StatePred<ClusterState>
                 &&& etcd_obj.metadata.resource_version.is_Some()
                 &&& etcd_obj.metadata.resource_version == req.obj.metadata.resource_version
                 &&& etcd_obj.metadata.owner_references.is_Some()
-                &&& exists |vrs: VReplicaSet| 
+                &&& exists |vrs: VReplicaSetView| 
                     #[trigger] owner_references.contains(vrs.controller_owner_ref())
             },
             // Dealt with similarly to update requests.
@@ -71,7 +71,7 @@ pub open spec fn vrs_not_interfered_by(other_id: int) -> StatePred<ClusterState>
                 &&& etcd_obj.metadata.resource_version.is_Some()
                 &&& etcd_obj.metadata.resource_version == req.obj.metadata.resource_version
                 &&& etcd_obj.metadata.owner_references.is_Some()
-                &&& exists |vrs: VReplicaSet| 
+                &&& exists |vrs: VReplicaSetView| 
                     #[trigger] owner_references.contains(vrs.controller_owner_ref())
             },
             // All delete requests to  carry a precondition on the resource version,
@@ -89,7 +89,7 @@ pub open spec fn vrs_not_interfered_by(other_id: int) -> StatePred<ClusterState>
                     &&& etcd_obj.metadata.resource_version
                         == req.preconditions.get_Some_0().resource_version
                     &&& etcd_obj.metadata.owner_references.is_Some()
-                    &&& exists |vrs: VReplicaSet| 
+                    &&& exists |vrs: VReplicaSetView| 
                         #[trigger] owner_references.contains(vrs.controller_owner_ref())
                 },
             _ => true,
--- a/src/v2/controllers/vreplicaset_controller/trusted/liveness_theorem.rs
+++ b/src/v2/controllers/vreplicaset_controller/trusted/liveness_theorem.rs
@@ -33,7 +33,6 @@ pub open spec fn owned_selector_match_is(vrs: VReplicaSetView, obj: DynamicObjec
     &&& obj.metadata.deletion_timestamp.is_None()
 }
 
-// TODO: the current not_interfered_by invariant is radically strong. Weaken it later.
 pub open spec fn vrs_not_interfered_by(other_id: int) -> StatePred<ClusterState> {
     |s: ClusterState| {
         forall |msg| {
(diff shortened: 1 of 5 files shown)


src/v2/controllers/vreplicaset_controller/trusted/liveness_theorem.rs
```diff
@@ -33,18 +33,63 @@ pub open spec fn owned_selector_match_is(vrs: VReplicaSetView, obj: DynamicObjec
...
         } ==> match msg.content.get_APIRequest_0() {
-            APIRequest::CreateRequest(req) => req.obj.kind != Kind::PodKind,
-            APIRequest::UpdateRequest(req) => req.obj.kind != Kind::PodKind,
-            APIRequest::UpdateStatusRequest(req) => req.obj.kind != Kind::PodKind,
-            APIRequest::DeleteRequest(req) => req.key.kind != Kind::PodKind,
+            // Other controllers don't create pods owned by a VReplicaSet.
+            APIRequest::CreateRequest(req) => !{
+                let owner_references = req.obj.metadata.owner_references.get_Some_0();
+                &&& req.obj.kind == Kind::PodKind
+                &&& req.obj.metadata.namespace.is_Some()
```
marshtompsxd, line 47: @codyjrivera do we need to require that requests from other controllers must carry a namespace? (tho they should)

src/v2/controllers/vreplicaset_controller/trusted/liveness_theorem.rs
```diff
@@ -33,18 +33,63 @@ pub open spec fn owned_selector_match_is(vrs: VReplicaSetView, obj: DynamicObjec
...
+                &&& req.obj.metadata.owner_references.is_Some()
+                &&& exists |i: int| 0 <= i < owner_references.len() ==>
+                        #[trigger] owner_references[i].kind == VReplicaSetView::kind()
```
marshtompsxd, line 50: @codyjrivera We also need to say that VRS is the **controller** owner reference which is unique. Consider using this `pod.metadata().owner_references_contains(v_replica_set.controller_owner_ref())` here: https://github.com/anvil-verifier/anvil/blob/main/src/v2/controllers/vreplicaset_controller/exec/reconciler.rs#L410

src/v2/controllers/vreplicaset_controller/trusted/liveness_theorem.rs
```diff
@@ -33,18 +33,63 @@ pub open spec fn owned_selector_match_is(vrs: VReplicaSetView, obj: DynamicObjec
...
+            APIRequest::UpdateRequest(req) => !{
+                let etcd_obj = s.resources()[req.key()];
+                let owner_references = etcd_obj.metadata.owner_references.get_Some_0();
+                &&& req.obj.kind == Kind::PodKind
+                &&& s.resources().contains_key(req.key())
+                &&& etcd_obj.metadata.resource_version.is_Some()
+                &&& etcd_obj.metadata.resource_version == req.obj.metadata.resource_version
+                &&& etcd_obj.metadata.owner_references.is_Some()
+                &&& exists |i: int| 0 <= i < owner_references.len() ==>
+                        #[trigger] owner_references[i].kind == VReplicaSetView::kind()
```
marshtompsxd, line 62: Same as above

src/v2/controllers/vreplicaset_controller/trusted/liveness_theorem.rs
```diff
@@ -33,18 +33,63 @@ pub open spec fn owned_selector_match_is(vrs: VReplicaSetView, obj: DynamicObjec
...
+            },
+            // Dealt with similarly to update requests.
+            APIRequest::UpdateStatusRequest(req) => !{
```
marshtompsxd, line 65: @codyjrivera can you put a TODO here saying that later we will allow other controllers to send UpdateStatus requests to owned pods after we address the fairness issues

src/v2/controllers/vreplicaset_controller/trusted/liveness_theorem.rs
```diff
@@ -33,18 +33,63 @@ pub open spec fn owned_selector_match_is(vrs: VReplicaSetView, obj: DynamicObjec
...
+                &&& etcd_obj.metadata.resource_version == req.obj.metadata.resource_version
+                &&& etcd_obj.metadata.owner_references.is_Some()
+                &&& exists |i: int| 0 <= i < owner_references.len() ==>
+                        #[trigger] owner_references[i].kind == VReplicaSetView::kind()
+            },
+            // All delete requests carry a precondition on the resource version,
+            // and other controllers don't try to delete pods owned by a VReplicaSet.
+            APIRequest::DeleteRequest(req) => 
+                req.preconditions.is_Some()
+                && req.preconditions.get_Some_0().resource_version.is_Some()
```
marshtompsxd, line 80: We don't have to require ALL delete requests to carry a precondition. Only the ones sent pods. You can write it as an implication.

src/v2/controllers/vreplicaset_controller/trusted/liveness_theorem.rs
```diff
@@ -33,18 +33,63 @@ pub open spec fn owned_selector_match_is(vrs: VReplicaSetView, obj: DynamicObjec
...
+                    let etcd_obj = s.resources()[req.key];
+                    let owner_references = etcd_obj.metadata.owner_references.get_Some_0();
+                    &&& req.key.kind == Kind::PodKind
+                    &&& s.resources().contains_key(req.key)
+                    &&& etcd_obj.metadata.resource_version.is_Some()
+                    &&& etcd_obj.metadata.resource_version
+                        == req.preconditions.get_Some_0().resource_version
+                    &&& etcd_obj.metadata.owner_references.is_Some()
+                    &&& exists |i: int| 0 <= i < owner_references.len() ==>
+                            #[trigger] owner_references[i].kind == VReplicaSetView::kind()
```
marshtompsxd, line 91: Same as above

src/v2/controllers/vreplicaset_controller/trusted/liveness_theorem.rs
codyjrivera: Not needed; thanks

src/v2/controllers/vreplicaset_controller/trusted/liveness_theorem.rs
codyjrivera: We don't have access to a singular `vrs` object in the non-interference property. We want the other controllers not to interfere with _any_ replicaset.

Perhaps
```
exists |vrs : VReplicaSetView| #[trigger] owner_references.contains(vrs.controller_owner_ref())
```
can express what we want here? The two statements are obviously not equivalent (look at the definition of controller_owner_ref, https://github.com/anvil-verifier/anvil/blob/45eb50563566d38b3347b82d0a1b37f410025791/src/v2/controllers/vreplicaset_controller/trusted/spec_types.rs#L23, to see that the former statement on lines 49-50 does not imply the latter definition I just gave).

Was that the gap you wanted me to plug?

marshtompsxd: > exists |vrs : VReplicaSetView| #[trigger] owner_references.contains(vrs.controller_owner_ref())

yes this should work

src/v2/controllers/vreplicaset_controller/trusted/liveness_theorem.rs
```diff
@@ -33,18 +33,65 @@ pub open spec fn owned_selector_match_is(vrs: VReplicaSetView, obj: DynamicObjec
...
+                &&& exists |vrs: VReplicaSetView| 
+                    #[trigger] owner_references.contains(vrs.controller_owner_ref())
+            },
+            // All delete requests to  carry a precondition on the resource version,
+            // and other controllers don't try to delete pods owned by a VReplicaSet.
+            APIRequest::DeleteRequest(req) => 
+                (req.key.kind == Kind::PodKind ==>
+                    req.preconditions.is_Some()
+                    && req.preconditions.get_Some_0().resource_version.is_Some())
+                && !{
```
marshtompsxd, line 83: nit: keep the indent consistent

marshtompsxd: @codyjrivera I think you should put everything as the consequent of `req.key.kind == Kind::PodKind ==>`

codyjrivera: Should be fixed. Also should have moved redundant kind comparisons

PR #608, opened 2025-04-21
Delete deprecated VRS controller

marshtompsxd:


Catoverflow (approved):

codyjrivera (approved):
It'd be nice, but low priority, to remove the v2 prefix elsewhere, since the `v2` prefix is already removed from the CI actions. Otherwise, I approve.

PR #611, opened 2025-04-22
Implement (simplified) Deployment controller and its model

Catoverflow:
- Add `VDeployment` controller model, implementation and documentation
- Add e2e test in CI and local test script for `VDeployment` controller
- As required for `VDeployment` controller, add
  - setter for `VReplicaSet` and its children fields
  - `Clone` trait for `VReplicaSet`
  - `well_formed` for `ObjectMeta` and `VReplicaSet`
  - `eq` for `PodTemplateSpec`
  - `Default` trait for `VReplicaSet`
  - picked `push_filter_and_filter_push` out of ad-hoc implementation from `VReplicaSet` to `vstd`
- Updated test scripts of `local-test.sh` and `deploy.sh` to separate logic for build and deploy (may be a bit out of scope)
 For e2e tests, I prefer `k8s.gcr.io/pause:3.9` over `php-redis`.
- Merge test scripts of admission controllers into `deploy.sh`


Catoverflow:
The current state machine implementation is hard for reasoning over termination. Should switch to another design which keeps track of old_vrs_list and new_vrs in state following the vanilla implementation.

src/kubernetes_api_objects/exec/object_meta.rs
```diff
@@ -186,6 +186,16 @@ impl ObjectMeta {
         self.inner.deletion_timestamp.is_some()
     }
 
+    #[verifier(external_body)]
+    pub fn well_formed(&self) -> (b: bool)
+        ensures b == self@.well_formed(),
+    {
+        self.inner.name.is_some()
```
codyjrivera, line 193: Super-ignorable nit: I'd personally indent this as?
```
self.inner.name.is_some()
&& self.inner.namespace.is_some()
&& self.inner.resource_version.is_some()
&& self.inner.uid.is_some()
```

src/v2/controllers/vdeployment_controller/exec/reconciler.rs
```diff
@@ -0,0 +1,505 @@
...
+        );
+        assert(
+            old_vrs_list@.map_values(|vrs: VReplicaSet| vrs@) ==
+            model_reconciler::filter_old_and_new_vrs(vrs_list@.map_values(|vrs: VReplicaSet| vrs@).take(0), vd@).1
+        );
+        assert(forall |i: int| 0 <= i < new_vrs_list.len() ==> new_vrs_list[i]@.well_formed());
+        assert(forall |i: int| 0 <= i < old_vrs_list.len() ==> old_vrs_list[i]@.well_formed());
+    }
+
+    while idx < vrs_list.len()
```
codyjrivera, line 378: Good job figuring out how to put both new and old lists into a single loop invariant!

src/v2/controllers/vdeployment_controller/exec/reconciler.rs
```diff
@@ -0,0 +1,505 @@
...
+    }
+    let mut labels = vd.spec().template().unwrap().metadata().unwrap().labels().unwrap().clone();
+    labels.insert("pod-template-hash".to_string(), pod_template_hash);
+    let mut label_selector = LabelSelector::default();
+    label_selector.set_match_labels(labels);
+    spec.set_selector(label_selector);
+    let template = template_with_hash(vd);
+    spec.set_template(template);
+    vrs.set_spec(spec);
+    proof {
```
codyjrivera, line 462: Is this `proof` block really needed anymore now we fixed the dash/underscore discrepancy?

src/v2/controllers/vdeployment_controller/exec/reconciler.rs
```diff
@@ -0,0 +1,505 @@
...
+{
+    let pod_template_hash = vd.metadata().resource_version().unwrap();
+    let mut labels = vd.spec().template().unwrap().metadata().unwrap().labels().unwrap();
+    labels.insert("pod-template-hash".to_string(), pod_template_hash);
+    let mut template_meta = ObjectMeta::default();
+    template_meta.set_labels(labels);
+    let mut pod_template_spec = PodTemplateSpec::default();
+    pod_template_spec.set_metadata(template_meta);
+    pod_template_spec.set_spec(vd.spec().template().unwrap().spec().unwrap());
+    proof {
```
codyjrivera, line 487: Similar to above

src/v2/controllers/vdeployment_controller/trusted/spec_types.rs
```diff
@@ -127,9 +124,11 @@ impl ResourceView for VDeploymentView {
         // template and its metadata ane spec exists
         &&& self.spec.template.is_Some()
         &&& self.spec.template.get_Some_0().metadata.is_Some()
+        // can be derived from selector match labels
+        &&& self.spec.template.get_Some_0().metadata.get_Some_0().labels.is_Some()
         &&& self.spec.template.get_Some_0().spec.is_Some()
         // selector matches template's metadata's labels
-        &&& self.spec.selector.matches(self.spec.template.get_Some_0().metadata.get_Some_0().labels.unwrap_or(Map::empty()))
+        &&& self.spec.selector.matches(self.spec.template.get_Some_0().metadata.get_Some_0().labels.get_Some_0())
```
codyjrivera, line 131: I'm confused about this line. Please explain.

codyjrivera (commented):
A few of my comments might be a bit stale given our meeting, fyi.

discussion/kubernetes-model/deployment_controller.md
```diff
@@ -0,0 +1,56 @@
...
+   ​	I. [rollout](https://github.com/kubernetes/kubernetes/blob/cdc807a9e849b651fb48c962cc18e25d39ec5edf/pkg/controller/deployment/rolling.go#L31): (Create if non-existing and) scale up new RS, then scale down old RS
+   ​	II. [recreate](https://github.com/kubernetes/kubernetes/blob/cdc807a9e849b651fb48c962cc18e25d39ec5edf/pkg/controller/deployment/recreate.go#L29): Scale down old RS, then (create if non-existing and) scale up the new RS.
+
+---
+
+*: The deployment controller will only control at most one new replica set, and multiple old replica sets
+
+> In k8s implementation it's [possible](https://github.com/kubernetes/kubernetes/blob/cdc807a9e849b651fb48c962cc18e25d39ec5edf/pkg/controller/deployment/util/deployment_util.go#L633-L634) to have multiple new replica sets in rare cases
+
+**: The scale up/down process should satisfy `maxSurge` and `maxUnavailable` properties of DC, which means DC may not scale the controlled new & old RS to desired replicas in one step. Instead DC will scale up & down gradually in turn. Currently we do not model this feature.
```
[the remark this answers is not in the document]
Catoverflow: Yes, let me update this ambiguous description here.

e2e/src/v2_vreplicaset_e2e.rs
```diff
@@ -16,6 +16,11 @@ use kube::{
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
+use deps_hack::anyhow::Result;
```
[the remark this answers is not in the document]
Catoverflow: No, fixed

src/v2/controllers/vdeployment_controller/trusted/spec_types.rs
Catoverflow: Should be accidentally introduced when I tried to have corresponding part in exec and failed. Reverted

Catoverflow:
@codyjrivera Thanks for your comments! I have updated some of them, while as I'm going to rewrite the VD some implementation-specific problems will be rechecked after the rewrite is complete.

src/v2/controllers/vdeployment_controller/model/reconciler.rs
```diff
@@ -0,0 +1,246 @@
...
+                        reconcile_step: VDeploymentReconcileStepView::RollReplicas,
+                        vrs_list: vrs_list,
+                        ..state
+                    };
+                    (state_prime, None)
+                }
+            }
+        },
+        VDeploymentReconcileStepView::RollReplicas => {
+            // TODO: support different policy (order of scaling of new and old vrs)
```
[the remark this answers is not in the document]
Catoverflow: I think the code makes decisions on which to update first is in this state, and it also determines how many to update (for `maxSurge` and `maxUnavailable`)

deploy.sh
```diff
@@ -15,8 +15,25 @@ app=$(echo "$1" | tr '_' '-') # should be the controller's name (with words sepa
...
+kind create cluster --config deploy/kind.yaml
+kind load docker-image local/$app-controller:v0.1.0 --name e2e
+
+# for VDeployment, need to deploy VReplicaSet as a dependency
+if [ "$app" == "v2-vdeployment" ]; then
+    kind load docker-image local/v2-vreplicaset-controller:v0.1.0 --name e2e
+fi
+
+# use imperative management for CRDs since metadata for PodTemplateSpec is too long.
+# why don't we directly create CRD using kube API?
```
marshtompsxd, line 35: @Catoverflow we don't want to assume what kube client lib people are using, so we use kubectl here.
btw if you want to leave a comment for future improvement, please start with `TODO: `

discussion/kubernetes-model/deployment_controller.md
marshtompsxd, line 1: @Catoverflow Thank you for this document. I suggest you move it to the code file (as comments). A problem of this type of doc is that they easily get out-of-synced when the code evolves. Putting the doc with the code does not solve the problem, but makes it better. It's also easier for anyone who reads the doc to relate to the code.

e2e/src/v2_vreplicaset_e2e.rs
```diff
@@ -16,6 +16,10 @@ use kube::{
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
+use deps_hack::serde_yaml;
+use deps_hack::tokio;
+use deps_hack::tracing::{error, info};
+use deps_hack::tracing_subscriber;
```
marshtompsxd, line 22: Do we need these new imports?

src/kubernetes_api_objects/exec/label_selector.rs
```diff
@@ -69,7 +70,9 @@ impl LabelSelector {
     // TODO: prove it and maybe move to a different lib
     #[verifier(external_body)]
     pub fn matches(&self, labels: StringMap) -> (res: bool)
-        ensures res == self@.matches(labels@)
+        ensures
+            res == self@.matches(labels@),
+            res ==> self@.match_labels.is_None() || labels@.len() >= self@.match_labels.get_Some_0().len(),
```
marshtompsxd, line 75: Why do we need this second postcondition here? Isn't it already implied by the first postcondition?

e2e/src/v2_vdeployment_e2e.rs
```diff
@@ -0,0 +1,496 @@
...
+    let discovery = Discovery::new(client.clone()).run().await?;
+    let vd_name = apply(v_deployment(), client.clone(), &discovery).await?;
+
+    desired_state_test(client.clone(), vd_name.clone()).await?;
+    scaling_test(client.clone(), vd_name.clone()).await?;
+    template_patch_test(client.clone(), vd_name.clone()).await?;
+
+    info!("E2e test passed.");
+    Ok(())
+}
```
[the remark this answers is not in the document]
Catoverflow: That is fine I guess? We separate tests in a "code block"

local-test.sh
```diff
@@ -3,18 +3,71 @@
 ## Test the controller locally in a kind cluster.
 ##
 ## Requires kind to be installed and the prerequisites of deploy.sh.
+## Usage: ./local-test.sh <controller_name> [--no-build]
```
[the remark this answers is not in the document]
Catoverflow: @codyjrivera I think the problem is `vargo` does not support build cache. So I can only reuse the binary as a whole

src/kubernetes_api_objects/exec/label_selector.rs
```diff
@@ -69,7 +70,9 @@ impl LabelSelector {
     // TODO: prove it and maybe move to a different lib
     #[verifier(external_body)]
     pub fn matches(&self, labels: StringMap) -> (res: bool)
-        ensures res == self@.matches(labels@)
+        ensures
+            res == self@.matches(labels@),
```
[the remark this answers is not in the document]
Catoverflow: @codyjrivera I think `labels@.len() >= self@.match_labels.get_Some_0().len()` is provable, and that's required by the `VReplicaSetView.state_validation()`, where `spec.selector` and its labels must be non-empty, and using this post-condition we know `template.metadata.labels` must be non-empty. But you are right, may be this should be written as a lemma?

deploy.sh
Catoverflow: After the discussion I think it's fine, let me remove it

src/kubernetes_api_objects/exec/label_selector.rs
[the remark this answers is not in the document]
Catoverflow: I tried and it proved without the lemma, should be fine. Let me remove it.

e2e/src/v2_vreplicaset_e2e.rs
Catoverflow: Removed

src/kubernetes_api_objects/exec/label_selector.rs
Catoverflow: It's required by `VReplicaSetView.state_validation` but it's automatically proved. Let me remove it.

Catoverflow:
@marshtompsxd I want to unify the test pipeline, but currently `v2-vreplicaset-admission` has its own scripts in `src/v2/controllers/vreplicaset_controller/admission_control`. I think it's better to move them to `deploy` and `e2e` so we don't need special treatment in ci and scripts. That's beyond the scope of this PR so I just added simple workaround in `deploy.sh` and changed the name in its scripts to use the same image name as other controllers

marshtompsxd:
> @marshtompsxd I want to unify the test pipeline, but currently `v2-vreplicaset-admission` has its own scripts in `src/v2/controllers/vreplicaset_controller/admission_control`. I think it's better to move them to `deploy` and `e2e` so we don't need special treatment in ci and scripts. That's beyond the scope of this PR so I just added simple workaround in `deploy.sh` and changed the name in its scripts to use the same image name as other controllers

I totally agree

discussion/kubernetes-model/deployment_controller.md
Catoverflow: I moved it to `src/v2/controllers/vdeployment_controller/trusted/step.rs`

src/vstd_ext/seq_lib.rs
```diff
@@ -271,6 +271,15 @@ pub proof fn push_to_set_eq_to_set_insert<A>(s: Seq<A>, e: A)
     }
 }
 
+pub proof fn push_filter_and_filter_push<A>(s: Seq<A>, pred: spec_fn(A) -> bool, e: A)
+    ensures
+        pred(e) ==> s.push(e).filter(pred) == s.filter(pred).push(e),
+        !pred(e) ==> s.push(e).filter(pred) == s.filter(pred),
+{
+    reveal(Seq::filter);
+    assert(s.push(e).drop_last() == s);
```
marshtompsxd, line 280: Just curious: is the last assertion here necessary?

src/v2/controllers/vreplicaset_controller/trusted/spec_types.rs
```diff
@@ -124,7 +138,8 @@ impl ResourceView for VReplicaSetView {
         &&& self.spec.template.get_Some_0().metadata.is_Some()
         &&& self.spec.template.get_Some_0().spec.is_Some()
         // selector matches template's metadata's labels
-        &&& self.spec.selector.matches(self.spec.template.get_Some_0().metadata.get_Some_0().labels.unwrap_or(Map::empty()))
+        &&& self.spec.template.get_Some_0().metadata.get_Some_0().labels.is_Some()
```
marshtompsxd, line 141: Just want to double check with you: the labels must be `Some`, is that correct?

src/v2/controllers/vreplicaset_controller/trusted/exec_types.rs
```diff
@@ -131,6 +174,19 @@ pub struct VReplicaSetSpec {
 impl VReplicaSetSpec {
     pub spec fn view(&self) -> spec_types::VReplicaSetSpecView;
 
+    #[verifier(external_body)]
+    pub fn default() -> (vreplicaset_spec: VReplicaSetSpec)
+        ensures vreplicaset_spec@ == spec_types::VReplicaSetSpecView::default(),
+    {
+        VReplicaSetSpec {
+            inner: deps_hack::VReplicaSetSpec {
```
marshtompsxd, line 182: Why not directly call `inner::default()`? VReplicaSetSpec has Default trait now https://github.com/anvil-verifier/anvil/pull/611/files#diff-4b77f89f66f064750dbbf3d4d2e54e55c175c6b36a5f2375c3ebe835f6f95e3bR275

src/kubernetes_api_objects/spec/object_meta.rs
```diff
@@ -80,6 +80,29 @@ impl ObjectMetaView {
         }
     }
 
+    pub open spec fn add_label(self, key: StringView, value: StringView) -> ObjectMetaView {
```
marshtompsxd, line 83: Why not call it `with_label`?

src/kubernetes_api_objects/exec/object_meta.rs
```diff
@@ -214,6 +224,28 @@ impl ObjectMeta {
...
+            let mut labels = std::collections::BTreeMap::new();
+            labels.insert(key, value);
+            self.inner.labels = Some(labels);
+        } else {
+            self.inner.labels.as_mut().unwrap().insert(key, value);
+        };
+    }
+
+    #[verifier(external_body)]
+    pub fn unset_label(&mut self, key: String)
```
marshtompsxd, line 241: How about calling it `remove_label`? Better to be consistent with `add_label`.

src/kubernetes_api_objects/exec/object_meta.rs
```diff
@@ -186,6 +186,16 @@ impl ObjectMeta {
         self.inner.deletion_timestamp.is_some()
     }
 
+    #[verifier(external_body)]
+    pub fn well_formed(&self) -> (b: bool)
+        ensures b == self@.well_formed(),
+    {
+        self.inner.name.is_some()
+            && self.inner.namespace.is_some()
```
marshtompsxd, line 194: K8s has cluster-wide objects which do not have a `namespace`, tho we currently do not support (and do not need) these objects. I would suggest you move `self.inner.namespace.is_some()` to each CR object's `well_formed`, and then you can leave a comment here to explain the decision.

deploy/v2_vdeployment/v2_vdeployment.yaml
```diff
@@ -0,0 +1,19 @@
...
+    matchLabels:
+      app: pause-demo
+  template:
+    metadata:
+      labels:
+        app: pause-demo
+    spec:
+      containers:
+      - name: pause
+        image: k8s.gcr.io/pause:3.9
```
marshtompsxd, line 19: What does `pause` do? Just sleep?

src/kubernetes_api_objects/exec/label_selector.rs
```diff
@@ -69,7 +70,9 @@ impl LabelSelector {
...
+            res ==> self@.match_labels.is_None(),
```
marshtompsxd, line 75: I think the second postcondition is wrong. "Matches" does not mean the labels is None

marshtompsxd: You should just remove this postcondition

src/v2/controllers/vdeployment_controller/model/reconciler.rs
```diff
@@ -0,0 +1,278 @@
...
+                reconcile_step: VDeploymentReconcileStepView::AfterGetReplicaSets,
+                new_vrs: None,
+                old_vrs_list: Seq::<VReplicaSetView>::empty(),
+            };
+            (state_prime, Some(RequestView::KRequest(req)))
+        },
+        VDeploymentReconcileStepView::AfterGetReplicaSets => {
+            if !(resp_o.is_Some() && resp_o.get_Some_0().is_KResponse()
+            && resp_o.get_Some_0().get_KResponse_0().is_ListResponse()
+            && resp_o.get_Some_0().get_KResponse_0().get_ListResponse_0().res.is_ok()) {
```
marshtompsxd, line 82: There is some shortcut macro you can use here: https://github.com/anvil-verifier/anvil/blob/main/src/v2/reconciler/spec/io.rs#L70

The macro for list is not defined yet. You will need to define it.

src/v2/controllers/vdeployment_controller/model/reconciler.rs
```diff
@@ -0,0 +1,278 @@
...
+            } else {
+                let objs = resp_o.unwrap().get_KResponse_0().get_ListResponse_0().res.unwrap();
+                let vrs_list_or_none = objects_to_vrs_list(objs);
+                if vrs_list_or_none.is_none() { // error in unmarshalling
+                    (error_state(state), None)
+                } else {
+                    let (new_vrs_list, old_vrs_list) = filter_old_and_new_vrs(filter_vrs_list(vrs_list_or_none.unwrap(), vd), vd);
+                    (
+                        VDeploymentReconcileState {
+                            reconcile_step: VDeploymentReconcileStepView::RollReplicas,
```
marshtompsxd, line 93: I assume that after you implement the second way of rolling, you are going to have a branch here and go to different next steps?

src/v2/controllers/vdeployment_controller/model/reconciler.rs
```diff
@@ -0,0 +1,278 @@
...
+                    new_vrs: Some(new_vrs),
+                    ..state
+                };
+                (state_prime, Some(RequestView::KRequest(req)))
+            } else {
+                // update existing new vrs
+                if !state.new_vrs.unwrap().well_formed() {
+                    (error_state(state), None)
+                } else {
+                    let new_vrs = state.new_vrs.get_Some_0(); // k8s deterministically chooses the "oldest" vrs, we just use the first one
```
marshtompsxd, line 121: Can you put a `TODO` here for `k8s deterministically chooses the "oldest" vrs, we just use the first one`?

src/v2/controllers/vdeployment_controller/model/reconciler.rs
```diff
@@ -0,0 +1,278 @@
...
+                    let existing_replicas = new_vrs.spec.replicas.unwrap_or(1);
```
marshtompsxd, line 122: Can you rename `existing_replicas` to `current_replicas`?

src/v2/controllers/vdeployment_controller/model/reconciler.rs
```diff
@@ -0,0 +1,278 @@
...
+                            },
+                            ..new_vrs
+                        };
+                        let req = APIRequest::UpdateRequest(UpdateRequest {
+                            name: new_vrs.metadata.name.unwrap(),
+                            namespace: namespace,
+                            obj: updated_new_vrs.marshal(),
+                        });
+                        let state_prime = VDeploymentReconcileState {
+                            reconcile_step: VDeploymentReconcileStepView::RollReplicas,
```
marshtompsxd, line 138: Why do we multiplex the `RollReplicas` step so much? Why not have different steps for managing new vrs and old vrs separately? The current way of representing steps makes it a bit hard to understand the state transition, and will create some difficulty in proving termination.

src/kubernetes_api_objects/exec/label_selector.rs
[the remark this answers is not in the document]
Catoverflow: It's indeed a legacy to be fixed. I fix it temporarily by changing the `vrs.state_validation`

src/v2/controllers/vdeployment_controller/model/reconciler.rs
Catoverflow: I tried and the macros are not usable because of the borrowed value. The correct version is 
```
        if !(resp_o.is_some() && resp_o.as_ref().unwrap().is_k_response()
            && resp_o.as_ref().unwrap().as_k_response_ref().is_list_response()
            && resp_o.as_ref().unwrap().as_k_response_ref().as_list_response_ref().res.is_ok()) {
```
Which is not covered by macro

Catoverflow:
@marshtompsxd I believe we have another [flaky error](https://github.com/anvil-verifier/anvil/actions/runs/14986228462/job/42100589351?pr=611) here for VReplicaSet:

```
error: assertion failed
    --> v2/controllers/vreplicaset_controller/proof/helper_invariants/proof.rs:1224:120
     |
1224 | ...ey(key) && stronger_next(s, s_prime) implies requirements(key, s_prime) by {
     |                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^ assertion failed
```

It's inside the proof of `lemma_eventually_always_each_vrs_in_reconcile_implies_filtered_pods_owned_by_vrs`. However, if I run manually:
```
verus -L dependency=src/deps_hack/target/debug/deps --extern=deps_hack="src/deps_hack/target/debug/libdeps_hack.rlib" --verify-module vreplicaset_controller::proof::helper_invariants::proof --verify-function lemma_eventually_always_each_vrs_in_reconcile_implies_filtered_pods_owned_by_vrs src/v2_vreplicaset_controller.rs --multiple-errors 100 --expand-errors
```
It succeeded with no error. @codyjrivera Can you help have a look at this? I think we encountered similar flaky proofs before.

codyjrivera:
Try `#[verifier(spinoff_prover)]` first. If that doesn't work, ping me and then I'll need to put debugging effort into it.

src/v2/controllers/vdeployment_controller/model/reconciler.rs
Catoverflow: Fixed in the most recent changes. Now we have different states for different uses.

src/v2/controllers/vdeployment_controller/model/reconciler.rs
Catoverflow: Yes, as commented now

src/kubernetes_api_objects/spec/object_meta.rs
Catoverflow: `with_labels` is for replacement, `add_label` and `remove_label` for incremental changes.

Catoverflow:
@codyjrivera It works. Thanks for your help! We can figure out how to simplify it later.

> Try #[verifier(spinoff_prover)] first. If that doesn't work, ping me and then someone will need to put debugging effort into it.

Catoverflow:
Finally this monster is ready for review again

src/v2/controllers/vdeployment_controller/trusted/exec_types.rs
marshtompsxd, line 1: @Catoverflow You can use https://github.com/anvil-verifier/anvil/blob/main/src/kubernetes_api_objects/exec/resource.rs#L86 to define VDeployment

src/v2/controllers/vdeployment_controller/trusted/liveness_theorem.rs
```diff
@@ -0,0 +1,28 @@
...
+    |s: ClusterState| {
+        let dyn_vrs_list = s.resources().values().to_seq().filter(|obj| obj.kind == VReplicaSetView::kind());
+        let vrs_list = objects_to_vrs_list(dyn_vrs_list);
+        let filtered_vrs_list = filter_vrs_list(vrs_list, vd);
+        let (new_vrs_list, old_vrs_list) = filtered_vrs_list;
+        &&& new_vrs_list.len() == 1
+        &&& old_vrs_list.len() == 0
+        &&& new_vrs_list[0].spec.replicas.unwrap_or(1) == vd.spec.replicas.unwrap_or(1)
+        &&& match_template_without_hash(vd, new_vrs_list[0])
+        &&& current_state_matches(new_vrs_list[0])
```
marshtompsxd, line 24: You can remove this last line

src/v2/controllers/vdeployment_controller/trusted/liveness_theorem.rs
```diff
@@ -0,0 +1,28 @@
...
+use crate::vdeployment_controller::trusted::spec_types::*;
+use crate::vdeployment_controller::model::reconciler::*;
+use crate::vreplicaset_controller::trusted::liveness_theorem::*;
+use vstd::prelude::*;
+
+verus !{
+
+// draft of ESR for VDeployment
+// TODO: add another version which talks about pods and derives from VRS ESR and this ESR
+pub open spec fn current_state_matches(vd: VDeploymentView) -> StatePred<ClusterState> {
```
marshtompsxd, line 14: Can you confirm that k8s deployment controller eventually deletes all old vrs? Or does it keep some old vrs but set their replicas to 0? If all old vrs are deleted eventually, you should just say vrs_list == Seq![one_new_vrs]; if not, you should say there is exactly one new vrs (whose template matches) and all other vrs' replicas are 0.

src/v2/controllers/vdeployment_controller/trusted/liveness_theorem.rs
```diff
@@ -0,0 +1,28 @@
+use crate::kubernetes_api_objects::spec::prelude::*;
+use crate::kubernetes_cluster::spec::cluster::*;
+use crate::temporal_logic::defs::*;
+use crate::vreplicaset_controller::trusted::spec_types::*;
+use crate::vdeployment_controller::trusted::spec_types::*;
+use crate::vdeployment_controller::model::reconciler::*;
```
marshtompsxd, line 6: Ideally, we should not import `crate::vdeployment_controller::model`. This file is the spec and should be read and understood by anyone using the controller, and `crate::vdeployment_controller::model` is part of the proof and only developers have to know about it.

src/kubernetes_api_objects/exec/object_meta.rs
```diff
@@ -186,6 +186,15 @@ impl ObjectMeta {
         self.inner.deletion_timestamp.is_some()
     }
 
+    #[verifier(external_body)]
+    pub fn well_formed(&self) -> (b: bool)
```
marshtompsxd, line 190: Thinking twice, could you name it as `well_formed_for_namespaced` and then get `self.metadata().namespace().is_some()` back to this function? I'm sorry for the back-and-forth requests!

src/v2/controllers/vreplicaset_controller/trusted/exec_types.rs
marshtompsxd, line 1: You can use https://github.com/anvil-verifier/anvil/blob/main/src/kubernetes_api_objects/exec/resource.rs#L86 here

src/v2/controllers/vdeployment_controller/trusted/spec_types.rs
```diff
@@ -127,9 +125,11 @@ impl ResourceView for VDeploymentView {
         // template and its metadata ane spec exists
         &&& self.spec.template.is_Some()
         &&& self.spec.template.get_Some_0().metadata.is_Some()
+        // can be derived from selector match labels
+        &&& self.spec.template.get_Some_0().metadata.get_Some_0().labels.is_Some()
```
marshtompsxd, line 129: I think deployment's template should not be option: https://github.com/Arnavion/k8s-openapi/blob/master/src/v1_30/api/apps/v1/deployment_spec.rs#L28

src/v2/controllers/vreplicaset_controller/trusted/spec_types.rs
```diff
@@ -152,4 +171,35 @@ pub struct VReplicaSetSpecView {
     pub template: Option<PodTemplateSpecView>,
```
marshtompsxd, line 171: This should not be Option (my bad for defining it this way). We should use this as reference: https://github.com/Arnavion/k8s-openapi/blob/master/src/v1_30/api/apps/v1/deployment_spec.rs#L28

src/v2/controllers/vdeployment_controller/trusted/liveness_theorem.rs
Catoverflow: But practically we will have 2 copies of utilities functions and need to prove the consistency between them like we once did for VRS, I'm weakly against this practice.

src/v2/controllers/vdeployment_controller/trusted/liveness_theorem.rs
Catoverflow: > Can you confirm that k8s deployment controller eventually deletes all old vrs?
It's configured in `d.Spec.RevisionHistoryLimit`, as the comment before `src/v2/controllers/vdeployment_controller/exec/reconciler.rs::reconcile_core` suggests

src/v2/controllers/vdeployment_controller/trusted/exec_types.rs
Catoverflow: I think these wrapper macros can reduce the duplication, but we still need to define other VDeployment-specific fields using the same structure. To me it's slightly better to use the same struct to define all methods as it improves readability

src/kubernetes_api_objects/exec/object_meta.rs
Catoverflow: I think it's better to put it in another PR, and keep the current `vd.well_formed` as is.

src/v2/controllers/vdeployment_controller/exec/reconciler.rs
```diff
@@ -0,0 +1,663 @@
...
+                return (error_state(state), None);
+            }
+            let objs = extract_some_k_list_resp!(resp_o).unwrap();
+            let vrs_list_or_none = objects_to_vrs_list(objs);
+            if vrs_list_or_none.is_none() {
+                return (error_state(state), None);
+            }
+            let (new_vrs_list, old_vrs_list) = filter_old_and_new_vrs(filter_vrs_list(vrs_list_or_none.clone().unwrap(), vd), vd);
+            // no .last().cloned() in verus because "The verifier does not yet support the following Rust feature: overloaded deref"
+            let state = VDeploymentReconcileState {
```
marshtompsxd, line 133: Here you bind the variable `state` to a different value than the original argument

src/v2/controllers/vdeployment_controller/exec/reconciler.rs
```diff
@@ -0,0 +1,663 @@
...
+                reconcile_step: VDeploymentReconcileStepView::Error,
+                new_vrs: None,
+                old_vrs_list: state.old_vrs_list@.map_values(|vrs: VReplicaSet| vrs@),
+            });
+            if new_vrs_list.len() == 0 {
+                // no new vrs, create one
+                let (state_prime, req) = create_new_vrs(state, vd);
+                assert(state_prime@ == model_reconciler::create_new_vrs(state@, vd@).0);
+                assert(req.deep_view() == model_reconciler::create_new_vrs(state@, vd@).1);
+                assert(state_prime@ == model_reconciler::reconcile_core(vd@, resp_o.deep_view(), state@).0) by {
```
marshtompsxd, line 148: ...and here you state that state_prime@ is the same as feeding the state@ to reconcile_core at model level. I don't think this assertion makes sense.

src/v2/controllers/vdeployment_controller/trusted/liveness_theorem.rs
marshtompsxd: You don't need to prove the consistency. You can just move them here, and let the model file import the liveness_theorem file. We don't have to do it in this PR.

src/v2/controllers/vdeployment_controller/exec/reconciler.rs
```diff
@@ -0,0 +1,644 @@
...
+        reconcile_done(state)
+    }
+
+    fn reconcile_error(state: &Self::S) -> bool {
+        reconcile_error(state)
+    }
+}
+
+#[verifier(external_body)]
+pub fn reconcile_init_state() -> (state: VDeploymentReconcileState)
```
marshtompsxd, line 65: Why isn't this one verified?

Catoverflow: Fixed, historical problem back that time.

error.log
marshtompsxd, line 1: Can you remove the error log?

Catoverflow:
Let me fix e2e test

marshtompsxd:
@Catoverflow Somehow the CI gets stuck even before starts. Can you make whatever no-op changes and push again to trigger CI?

Catoverflow:
@marshtompsxd CI is run but not showed in this PR page. The commit number and branch name are correct.
see https://github.com/anvil-verifier/anvil/actions/runs/15050205275

deploy/v2_vdeployment/v2_vdeployment.yaml
Catoverflow: Yes

PR #615, opened 2025-04-29
Repair all proofs broken by VRS Non-Interference Rely Condition

codyjrivera:
I also make a few changes to the rely-condition:
- For the Create portion, it actually suffices to just ensure that the object in the create message isn't controller-owned by any VRS.
- For the Update and UpdateStatus portions, we require that those requests carry a resource version. We need this because the current state machine model actually allows unconditional updates on `Pods` (but disallows this for CustomResources).

### DIFF
--- a/src/v2/controllers/vreplicaset_controller/trusted/liveness_theorem.rs
+++ b/src/v2/controllers/vreplicaset_controller/trusted/liveness_theorem.rs
@@ -73,7 +73,8 @@ pub open spec fn vrs_not_interfered_by_update_req(req: UpdateRequest) -> StatePr
     }
 }
 
-// Dealt with similarly to update requests.
+// Dealt with similarly to update requests, minus the condition on 
+// owner_references.
 // TODO: allow other controllers to send UpdateStatus
 // requests to owned pods after we address the fairness issues.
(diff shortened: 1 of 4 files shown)


src/v2/controllers/vreplicaset_controller/trusted/liveness_theorem.rs
```diff
@@ -37,31 +37,34 @@ pub open spec fn owned_selector_match_is(vrs: VReplicaSetView, obj: DynamicObjec
...
-            &&& s.resources().contains_key(req.key())
-            &&& etcd_obj.metadata.resource_version.is_Some()
-            &&& etcd_obj.metadata.resource_version == req.obj.metadata.resource_version
-            &&& etcd_obj.metadata.owner_references.is_Some()
-            &&& exists |vrs: VReplicaSetView| 
-                #[trigger] owner_references.contains(vrs.controller_owner_ref())
-        }
+        req.obj.kind == Kind::PodKind ==>
+            req.obj.metadata.resource_version.is_Some()
+            && (req.obj.metadata.owner_references.is_Some() ==>
```
marshtompsxd, line 55: @codyjrivera could you add more comments to explain why we want this condition: we want to prevent any other from updating a pod (whose controller owner is not vrs) so that its controller owner becomes vrs?

PR #616, opened 2025-05-01
Specify guarantee-condition for VReplicaSet controller

codyjrivera:
I add a guarantee-condition for `VReplicaSet`, constraining the messages sent to be create or delete requests on pods owned by the vrs controller.

I also do some cosmetic fixes here as well, changing the naming scheme from 'non-interference' to 'rely-guarantee'.


src/v2/controllers/vreplicaset_controller/trusted/rely_guarantee.rs
```diff
@@ -0,0 +1,167 @@
...
+// VRS Guarantee Condition
+
+// VRS only creates pods owned by a VReplicaSet.
+pub open spec fn vrs_guarantee_create_req(req: CreateRequest) -> StatePred<ClusterHistory> {
+    |h: ClusterHistory| {
+        let owner_references = req.obj.metadata.owner_references.get_Some_0();
+        &&& req.obj.kind == Kind::PodKind
+        &&& req.obj.metadata.owner_references.is_Some()
+        &&& exists |vrs: VReplicaSetView| 
+            #[trigger] owner_references.contains(vrs.controller_owner_ref())
```
marshtompsxd, line 122: @codyjrivera You might want to enhance it to state that there is only one controller_owner_ref in owner_references which is vrs.

src/v2/controllers/vreplicaset_controller/trusted/rely_guarantee.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+}
+
+// VRS only sends delete requests to pods known to have been in etcd and 
+// owned by a VReplicaSet at some point in history. The request will
+// carry a precondition that matches the resource version of the historical
+// etcd object.
+pub open spec fn vrs_guarantee_delete_req(req: DeleteRequest) -> StatePred<ClusterHistory> {
```
marshtompsxd, line 130: The delete guarantee should not talk about the history. It should just say for any pod delete request sent by this vrs controller, if that object in etcd matches the RV in precondition, then that object is owned by VRS.

marshtompsxd: You will have intermediate lemmas that talk about history, but they are only used for proof. They should not be part of the interface.

src/v2/controllers/vreplicaset_controller/trusted/rely_guarantee.rs
codyjrivera: this is guaranteed by an init invariant.

marshtompsxd: Yes, `each_object_in_etcd_has_at_most_one_controller_owner` can do the job, but then you will not be able to prove that VRS's guarantee implies another controller's rely --- you can only prove that with VRS's guarantee condition and this invariant, VRS's creation request will not interfere with another controller.

Ideally, you could structure the compositional proof in a way that (1) first prove A's guarantee implies B's rely, and (2) then prove B's liveness under B's rely.

The proof of (2) will be much more complex than (1), so you only want to do it once and it should not depend on A. The proof of (1) should be much easier and you will need to do it for each pair of controllers.

marshtompsxd:
Ready for review again?

src/v2/controllers/vreplicaset_controller/trusted/rely_guarantee.rs
```diff
@@ -0,0 +1,164 @@
...
+                == req.preconditions.get_Some_0().resource_version
+        } ==> {
+            let controller_owners = etcd_obj.metadata.owner_references
+                .get_Some_0()
+                .filter(|o: OwnerReferenceView| {
+                    o.controller.is_Some() && o.controller.get_Some_0()
+                });
+            &&& etcd_obj.metadata.owner_references.is_Some()
+            &&& exists |vrs: VReplicaSetView| 
+                controller_owners == seq![#[trigger] vrs.controller_owner_ref()]
```
marshtompsxd, line 143: This seems too strong? The pod VRS controller deletes might have more than one owners, but only one controller owner (which is VRS)

marshtompsxd:
@codyjrivera in addition to my new comment, can you also update the description to be more concise?

src/v2/controllers/vreplicaset_controller/trusted/rely_guarantee.rs
codyjrivera: Look again; `controller_owners` is filtered precisely that way.

codyjrivera:
Updated the PR description; if you meant any comments in the code, please point them out to me (I think the existing ones are okay).

src/v2/controllers/vreplicaset_controller/trusted/rely_guarantee.rs
marshtompsxd: I double checked the implementation of the reference RS controller: it manages pods whose controller owner is RS, even if the pods have other owners. See: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/controller_ref_manager.go#L70-L75

codyjrivera: Perhaps I should have explained further originally. `controller_owners` is defined precisely as the subset of `owner_references` that are controller owners. The line enforces that the only _controller_ owner is a VReplicaSet. My definition allows other non-controller owners.

PR #619, opened 2025-05-03
Make docker builder optional, and simplify local test step

Catoverflow:
This PR makes test workflow easier to run locally.

Changes:
- Add Rust toolchain to cache
- Add build option without docker builder and update existing CI workflows
- Update `local-test.sh` with support for local builds
- Add documentation in `build.md`
- Update builder and controller base image to Ubuntu 22.04, same as Github CI environment. New builder image is already pushed.

The reason I want to deprecate the docker builder is `vargo` does not reuse cache and include some changes in built binary each run (even with identical source), and that causes docker to build a new image from start, which requires a new builder image. The problem is worse with vdeployment controller, which depends on vreplicaset controller, and result in 4 fresh-built docker images (`(vd+vrs) * (builder+controller)`) per run locally. This quickly eats up my disk and time and I have to run `docker prune` every time. For cases where builder is required (different runtime environment between host and controller base image), builder can still be used. Detailed usage is in `build.md`.


Catoverflow:
@agrama64 I'm deprecating the docker builder, and use image with locally built controller in this PR. However, the admission ci broke and I didn't figure out how to fix. Sorry but could you have a look at [this job](https://github.com/anvil-verifier/anvil/actions/runs/14815587620/job/41595562952)?

build.md
```diff
@@ -0,0 +1,73 @@
...
+   Base image and builder image is specified in `docker/controller/Dockerfile.[local|remote]` respectively
+3. Setup cluster, apply controller image using [kind](https://kind.sigs.k8s.io/).
+4. Apply test specified in `e2e/src` and workload in `deploy` by `deploy.sh`
+
+This process can be automated with
+
+```
+./local-test.sh <controller_name> [--build|--build-remote]
+Usage:
+	--build:			Call ./build.sh to build the controller before test, should have VERUS_DIR speccified
```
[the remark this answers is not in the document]
Catoverflow: @marshtompsxd From the view of typora it's aligned. Sometimes there are inconsistencies as we don't have a spec for markdown. Do you want me to follow Github's rendering result?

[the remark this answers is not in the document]
Catoverflow: Fixed

.github/workflows/ci.yml
marshtompsxd, line 1: What are these changes for?

marshtompsxd: oh I just realized that each task in the previous version of CI does not skip installing the rust toolchain even if it's based on `build-verus`. Is that the problem?

marshtompsxd:
I don't understand why you name it as "Deprecate docker builder...". I think you are still using the docker builder, no? Please correct me if my understanding is wrong.

.github/workflows/ci.yml
Catoverflow: Yes, I cached Rust toolchain together with Verus, then it's unnecessary.

Catoverflow:
> I don't understand why you name it as "Deprecate docker builder...". I think you are still using the docker builder, no? 
Yes I forgot to update the title...it's now completed

PR #621, opened 2025-05-10
Support "transactional" Get-Then-Update to address fairness issues on version races

marshtompsxd:


marshtompsxd:
@codyjrivera @Catoverflow Sorry that this PR accidentally contains some changes for formatting the proof of VRS controllers (mostly on removing whitespace); I turned on format-on-save in my dev env so...

src/v2/shim_layer/controller_runtime.rs
```diff
@@ -385,6 +408,157 @@ where
     return Ok(Action::requeue(Duration::from_secs(60)));
 }
 
+// transactional_get_then_delete_by_retry retries get and then delete upon conflict errors to simulate atomic operations.
```
codyjrivera, line 411: More of a basic Kubernetes question than a comment on this code, but do real-life examples of this 'read->update' loop have a timeout, or would they just contend for a resource forever?

src/v2/kubernetes_cluster/spec/api_server/state_machine.rs
```diff
@@ -652,6 +651,90 @@ pub open spec fn handle_update_status_request_msg(installed_types: InstalledType
...
+    let get_req = GetRequest {
+        key: req.key()
+    };
+    let get_resp = handle_get_request(get_req, s);
+    match get_resp.res {
+        Ok(_) => {
+            let current_obj = s.resources[req.key()];
+            // Step 2: if the object exists, perform a check using a predicate on object
+            // The predicate: Is the current object owned by req.owner_ref?
+            // TODO: the predicate should be provided by clients instead of the hardcoded one
```
codyjrivera, line 711: Long shot: might there be a way to specify this predicate at compile/typechecking time, rather than either hardcoding it or trying to pass in a higher-order function?

codyjrivera (commented):
Looks good. Obviously the predicate being hardcoded isn't ideal, but good work!

Ideally, more people should check the mathematics of course.

src/v2/shim_layer/controller_runtime.rs
marshtompsxd: They limit the number of retries (and also have back-off when a try fails).

src/v2/kubernetes_cluster/spec/api_server/state_machine.rs
marshtompsxd: I hope there is a way without using `dyn`. Need to ask @utaal

src/v2/shim_layer/controller_runtime.rs
marshtompsxd: See this https://github.com/kubernetes/kubernetes/blob/0e64c6443f8e1f760c92a64304925986d4519a77/staging/src/k8s.io/client-go/util/retry/util.go#L68

codyjrivera: Is there some other way that would be facilitated? Or are we content, as a simplifier for the purposes of Anvil, with possible starvation?

marshtompsxd: If we limit the number of retries then there is no way to have a refinement mapping from the retry code to the transactional spec API.

We could implement back-off later, which does not affect the liveness argument.

> with possible starvation?

What do you mean by starvation here? You mean the controller always retries and never terminates?

codyjrivera: Understandable, I was just curious.

I meant 'always retries'.

src/v2/kubernetes_cluster/spec/api_server/state_machine.rs
Catoverflow: I think we need to have lemma like `lemma_always_key_of_object_in_matched_ok_create_resp_message_is_same_as_key_of_pending_req`, do we want to add those in another PR and then we can migrate to use this API?

PR #625, opened 2025-05-13
Prove VReplicaSet Guarantee Condition

codyjrivera:
This pull request introduces the proof of VRS's guarantee condition, thus completing the _local_ rely-guarantee obligations for our verified ReplicaSet.

Notably, I was able to prove the guarantee condition without the use of history variables. In fact, much of the legwork comes from the invariant `each_vrs_in_reconcile_implies_filtered_pods_owned_by_vrs`, which I was able to modify to prove without weak fairness and only with invariants starting from `init` (rather than those only true at a later point in the trace).


src/v2/controllers/vreplicaset_controller/proof/liveness/proof.rs
marshtompsxd, line 1: What are these changes for?

src/v2/controllers/vreplicaset_controller/proof/guarantee.rs
```diff
@@ -0,0 +1,258 @@
...
+        rely_guarantee::*,
+        spec_types::*, 
+    },
+};
+use crate::vstd_ext::seq_lib::*;
+use vstd::prelude::*;
+
+verus! {
+
+spec fn stronger_vrs_guarantee_delete_req(req: DeleteRequest) -> StatePred<ClusterState> {
```
marshtompsxd, line 18: Could you add a comment to explain this invariant?

src/v2/controllers/vreplicaset_controller/proof/liveness/proof.rs
codyjrivera: Phase IV served only to prove the invariant `each_vrs_in_reconcile_implies_filtered_pods_owned_by_vrs`. I was able to prove this invariant (and modify it slightly) so that it held starting from the beginning. Thus, Phase IV became redundant, and I deleted it.

src/v2/controllers/vreplicaset_controller/proof/guarantee.rs
codyjrivera: I improved the comment.

marshtompsxd:
Seems that there is a flaky proof?

marshtompsxd:
I find that vreplicaset-admission-verification is purely redundant; there is nothing to verify there... Let me remove it

PR #630, opened 2025-05-15
Specify non interference rely/guarantee conditions of deployment controller

Catoverflow:
- add reliance conditions for deployment controller
- add TODO for type invariants in transactional API

### DIFF
--- a/src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
+++ b/src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
@@ -58,14 +58,14 @@ pub open spec fn vd_rely_get_then_update_req(req: GetThenUpdateRequest) -> State
                 // req.owner_ref.controller.is_Some_and(true) && the same as vd.controller_owner_ref()
                 &&& exists |vd: VDeploymentView| req.owner_ref == vd.controller_owner_ref()
                 // is one of the owners of vrs
-                &&& s.resources().contains_key(req.key)
-                &&& #[trigger] s.resources()[req.key].metadata.owner_references_contains(req.owner_ref)
+                &&& s.resources().contains_key(req.key())
+                &&& #[trigger] s.resources()[req.key()].metadata.owner_references_contains(req.owner_ref)
             }
             // Prevents 2): where other controllers update vrs so they become
             // owned by a VDeployment.
             && (req.obj.metadata.owner_references.is_Some() ==>
                 forall |vd: VDeploymentView| 
-                    ! #[trigger] req.obj.metadata.owner_references.get_Some_0().contains(vd.controller_owner_ref()))
+                    ! req.obj.metadata.owner_references.get_Some_0().contains(vd.controller_owner_ref()))
 
     }
 }
@@ -116,7 +116,7 @@ pub open spec fn vd_rely_get_then_delete_req(req: GetThenDeleteRequest) -> State
                 &&& exists |vd: VDeploymentView| req.owner_ref == vd.controller_owner_ref()
                 // is one of the owners of vrs
                 &&& s.resources().contains_key(req.key)
-                &&& #[trigger] s.resources()[req.key].metadata.owner_references_contains(req.owner_ref)
+                &&& s.resources()[req.key].metadata.owner_references_contains(req.owner_ref)
             }
     }
 }
@@ -126,7 +126,7 @@ pub open spec fn vd_rely(other_id: int) -> StatePred<ClusterState> {
         forall |msg| {
             &&& #[trigger] s.in_flight().contains(msg)
             &&& msg.content.is_APIRequest()
-            &&& msg.src == HostId::Controller(other_id)
+            &&& msg.src.is_controller_id(other_id)
         } ==> match msg.content.get_APIRequest_0() {
             APIRequest::CreateRequest(req) => vd_rely_create_req(req)(s),
             APIRequest::UpdateRequest(req) => vd_rely_update_req(req)(s),
@@ -177,7 +177,7 @@ pub open spec fn vd_guarantee(controller_id: int) -> StatePred<ClusterState> {
         forall |msg| {
             &&& #[trigger] s.in_flight().contains(msg)
             &&& msg.content.is_APIRequest()
-            &&& msg.src == HostId::Controller(controller_id)
+            &&& msg.src.is_controller_id(controller_id)
         } ==> match msg.content.get_APIRequest_0() {
             APIRequest::CreateRequest(req) => vd_guarantee_create_req(req)(s),
             APIRequest::GetThenUpdateRequest(req) => vd_guarantee_get_then_update_req(req)(s),
--- a/src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
+++ b/src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
@@ -54,7 +54,7 @@ pub open spec vd_rely_update_req(req: UpdateRequest) -> StatePred<ClusterState> 
 // Q: how to get controller type in ClusterState?
 pub open spec fn vd_rely_update_status_req(req: UpdateStatusRequest) -> StatePred<ClusterState> {
     |s: ClusterState| {
-        req.obj.kind == Kind::CustomResourceKind("vreplicaset"@) ==> 
+        req.obj.kind == Kind::CustomResourceKind("vdeployment"@) ==> 
             req.obj.metadata.resource_version.is_Some()
             && !{
                 let etcd_obj = s.resources()[req.key()];
@@ -86,7 +86,7 @@ pub open spec fn vd_rely_delete_req(req: DeleteRequest) -> StatePred<ClusterStat
     }
 }
 
-pub open spec fn vd_reply(other_id: int) -> StatePred<ClusterState> {
+pub open spec fn vd_rely(other_id: int) -> StatePred<ClusterState> {
     |s: ClusterState| {
         forall |msg| {
             &&& #[trigger] s.in_flight().contains(msg)
@@ -131,7 +131,7 @@ pub open spec fn vd_guarantee_update_req(req: UpdateRequest) -> StatePred<Cluste
 //      and will not delete when the number of vrs is not greater than vd.Spec.RevisionHistoryLimit
 pub open spec fn vd_guarantee_delete_req(req: DeleteRequest) -> StatePred<ClusterState> {
     |s: ClusterState| {
-        true
+        false
     }
 }
 
--- a/src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
+++ b/src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
@@ -2,6 +2,7 @@ use crate::kubernetes_api_objects::spec::prelude::*;
 use crate::kubernetes_cluster::spec::{cluster::*, message::*};
 use crate::temporal_logic::defs::*;
 use crate::vdeployment_controller::trusted::spec_types::*;
+use crate::vreplicaset_controller::trusted::spec_types::*;
 use vstd::prelude::*;
 
 verus! {
@@ -24,7 +25,7 @@ pub open spec fn vd_rely_create_req(req: CreateRequest) -> StatePred<ClusterStat
 // and cannot 1) update vrs owned by a VDeployment
 //         or 2) update vrs to become owned
 // by a VDeployment.
-pub open spec vd_rely_update_req(req: UpdateRequest) -> StatePred<ClusterState> {
+pub open spec fn vd_rely_get_then_update_req(req: GetThenUpdateRequest) -> StatePred<ClusterState> {
     |s: ClusterState| {
         req.obj.kind == Kind::CustomResourceKind("vreplicaset"@) ==>
             req.obj.metadata.resource_version.is_Some()
@@ -37,6 +38,8 @@ pub open spec vd_rely_update_req(req: UpdateRequest) -> StatePred<ClusterState> 
                 &&& etcd_obj.metadata.resource_version.is_Some()
                 &&& etcd_obj.metadata.resource_version == req.obj.metadata.resource_version
                 &&& etcd_obj.metadata.owner_references.is_Some()
+                // Can pass the owner_references check
+                &&& owner_references.contains(req.owner_ref)
                 &&& exists |vd: VDeploymentView| 
                     #[trigger] owner_references.contains(vd.controller_owner_ref())
             }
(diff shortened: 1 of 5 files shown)


src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
```diff
@@ -0,0 +1,153 @@
...
+                    forall |vd: VDeploymentView| 
+                        ! #[trigger] req.obj.metadata.owner_references.get_Some_0().contains(vd.controller_owner_ref()))
+    }
+}
+
+// TODO: double check if only vrs can send update status req to itself
+//       the update to VRS's guarantee to support this should also be updated
+// similar to update requests, minus the condition on owner_references.
+// Q: how to get controller type in ClusterState?
+pub open spec fn vd_rely_update_status_req(req: UpdateStatusRequest) -> StatePred<ClusterState> {
```
marshtompsxd, line 55: Is this rely condition really needed? It doesn't matter if some controller wants to update the `status` of vreplicaset objects. On the other hand, the vd controller does not want other controllers to update `status` of vd objects.

src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+}
+
+// TODO: support delete requests (not yet implemented in the controller)
+// This should be the same as vrs_guarantee_delete_req
+// plus vd will not delete new vrs (has same template as vd)
+//      and will not delete when the number of vrs is not greater than vd.Spec.RevisionHistoryLimit
+pub open spec fn vd_guarantee_delete_req(req: DeleteRequest) -> StatePred<ClusterState> {
+    |s: ClusterState| {
+        true
```
marshtompsxd, line 134: This should be `false` before you implement deletion

src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
```diff
@@ -0,0 +1,153 @@
...
+                &&& etcd_obj.metadata.resource_version
+                    == req.preconditions.get_Some_0().resource_version
+                &&& etcd_obj.metadata.owner_references.is_Some()
+                &&& exists |vrs: VReplicaSetView| 
+                    #[trigger] owner_references.contains(vrs.controller_owner_ref())
+            }
+    }
+}
+
+pub open spec fn vd_reply(other_id: int) -> StatePred<ClusterState> {
```
marshtompsxd, line 89: typo: rely

marshtompsxd:
@Catoverflow After #632 , we need to update the rely and guarantee conditions to talk about transaction APIs

src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
Catoverflow: Yes, this was a draft and I forgot to add the condition.

src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
```diff
@@ -0,0 +1,159 @@
...
+                &&& etcd_obj.metadata.resource_version.is_Some()
+                &&& etcd_obj.metadata.resource_version == req.obj.metadata.resource_version
+            }
+    }
+}
+
+// Other controllers don't try to delete vrs owned by a VDeployment.
+pub open spec fn vd_rely_get_then_delete_req(req: GetThenDeleteRequest) -> StatePred<ClusterState> {
+    |s: ClusterState| {
+        req.key.kind == Kind::CustomResourceKind("vreplicaset"@) ==>
```
codyjrivera, line 74: Nit: Can we import the VReplicaSet kind and do `VReplicaSetView::kind()` instead?

src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
```diff
@@ -0,0 +1,159 @@
...
+
+// VD only updates VRS owned by itself.
+pub open spec fn vd_guarantee_get_then_update_req(req: GetThenUpdateRequest) -> StatePred<ClusterState> {
+    |s: ClusterState| {
+        let etcd_obj = s.resources()[req.key()];
+        &&& req.obj.kind == Kind::CustomResourceKind("vreplicaset"@)
+        &&& req.obj.metadata.resource_version.is_Some()
+        &&& exists |vd: VDeploymentView|
+            req.owner_ref == vd.controller_owner_ref()
+            && req.obj.metadata.owner_references.get_Some_0() == seq![#[trigger] vd.controller_owner_ref()]
```
codyjrivera, line 130: Objects can have multiple _non_ controller owners, but only one controller owner. See vrs's `rely_guarantee.rs` (in `../../vreplicaset_controller/trusted`) for more inspiration.

src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+}
+
+// TODO: support delete requests (not yet implemented in the controller)
+// This should be the same as vrs_guarantee_delete_req
+// plus vd will not delete new vrs (has same template as vd)
+//      and will not delete when the number of vrs is not greater than vd.Spec.RevisionHistoryLimit
+pub open spec fn vd_guarantee_get_then_delete_req(req: GetThenDeleteRequest) -> StatePred<ClusterState> {
```
codyjrivera, line 138: I'd ordinarily suggest not to stub it out, but I think this is good to document what we want to do.

src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
```diff
@@ -0,0 +1,159 @@
...
+    |s: ClusterState| {
+        false
+    }
+}
+
+pub open spec fn vd_guarantee(other_id: int) -> StatePred<ClusterState> {
```
codyjrivera, line 144: This is a problem in `vrs_guarantee` too, but it should be `controller_id` rather than `other_id` (since we're describing _our_ messages).

codyjrivera (commented):
In the rely condition, we still need to handle `UpdateRequest` and `DeleteRequest` (unless those were removed from the cluster entirely, which reading Xudong's PR, they weren't).

src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
marshtompsxd: +1

marshtompsxd:
> In the rely condition, we still need to handle `UpdateRequest` and `DeleteRequest` (unless those were removed from the cluster entirely, which reading Xudong's PR, they weren't).

Yes

src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
marshtompsxd: + 1

src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
```diff
@@ -0,0 +1,205 @@
...
+    }
+}
+
+// TODO: double check if only vrs can send update status req to itself
+//       the update to VRS's guarantee to support this should also be updated
+// similar to update requests, minus the condition on owner_references.
+// Q: how to get controller type in ClusterState?
+pub open spec fn vd_rely_update_status_req(req: UpdateStatusRequest) -> StatePred<ClusterState> {
+    |s: ClusterState| {
+        req.obj.kind == Kind::CustomResourceKind("vdeployment"@) ==>
```
codyjrivera: It would be nice for you to do something similar for `VDeployment::view()`

src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
```diff
@@ -0,0 +1,205 @@
...
+
+// VD only updates VRS owned by itself.
+pub open spec fn vd_guarantee_get_then_update_req(req: GetThenUpdateRequest) -> StatePred<ClusterState> {
+    |s: ClusterState| {
+        let etcd_obj = s.resources()[req.key()];
+        &&& req.obj.kind == VReplicaSetView::kind()
+        &&& req.obj.metadata.resource_version.is_Some()
+        &&& exists |vd: VDeploymentView|
+            req.owner_ref == vd.controller_owner_ref()
+            && req.obj.metadata.owner_references_contains(vd.controller_owner_ref())
```
codyjrivera, line 176: On the other hand, this is too weak; we need to ensure that `vd.controller_owner_ref()` is the only _controller_ owner reference.

You'll need to come up with a 'compromise' by filtering `metadata.owner_references`.

codyjrivera: (I got these same pair of comments from Xudong a while ago btw!)

Catoverflow: Yes, I saw the complex conditions in VRS to ensure there's only one controller owner

codyjrivera: It doesn't appear to be implemented here.

src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
```diff
@@ -0,0 +1,201 @@
...
+                    .filter(|o: OwnerReferenceView| {
+                        o.controller.is_Some() && o.controller.get_Some_0()
+                    });
+                &&& s.resources().contains_key(req.key)
+                &&& etcd_obj.metadata.owner_references_contains(req.owner_ref)
+                &&& exists |vd: VDeploymentView| 
+                    controller_owners == seq![#[trigger] vd.controller_owner_ref()]
+            }
+    }
+}
```
marshtompsxd, line 133: This one could be simplified: it's sufficient to say that (1) `req.owner_ref`'s `controller` field is true and (2) `req.owner_ref` is not a `vd.controller_owner_ref()`. Since each object in etcd has only one controller owner ref, this two conditions means that such get-then-delete requests will not delete any VRS objects controllerly owned by VD.

src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
```diff
@@ -0,0 +1,201 @@
...
+            // by a VDeployment.
+            && !{
+                let etcd_obj = s.resources()[req.key()];
+                let owner_references = etcd_obj.metadata.owner_references.get_Some_0();
+                &&& s.resources().contains_key(req.key())
+                &&& etcd_obj.metadata.owner_references.is_Some()
+                // Can pass the owner_references check
+                &&& owner_references.contains(req.owner_ref)
+                &&& exists |vd: VDeploymentView| 
+                    #[trigger] owner_references.contains(vd.controller_owner_ref())
```
marshtompsxd, line 67: Similar comments to get-then-delete rely condition for code above

src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
Catoverflow: > each object in etcd has only one controller owner ref

Where should we place this spec? My worry is if we decouple this spec from the current rely condition, we also encode implicit knowledge about cluster in reliance conditions.

Catoverflow: And I think 
```
req.owner_ref.is_Some() && req.owner_ref.get_Some_0().controller &&
    req.owner_ref != vd.controller_owner_ref()
```
is too strong, maybe
```
req.owner_ref.is_Some() && req.owner_ref.get_Some_0().controller ==>
    req.owner_ref != vd.controller_owner_ref()
```
is better

marshtompsxd: The goal is to prevent other controllers from updating/deleting VRS objects whose controller owner is a VD.
```
req.owner_ref.is_Some() && req.owner_ref.get_Some_0().controller ==>
    req.owner_ref != vd.controller_owner_ref()
```
How does this condition prevent it? Note that an object could have multiple owners (and only one is a controller owner).

marshtompsxd: Ideally, I'd like to use type system to enforce that `req.owner_ref` is a controller owner, but that does not seem natural in Verus. So from the type perspective, `req.owner_ref`'s `controller` field might be false, but what we want to enforce is something stronger.

Catoverflow: > ```
> req.owner_ref.is_Some() && req.owner_ref.get_Some_0().controller ==>
>     req.owner_ref != vd.controller_owner_ref()
>     ```

I think this is stronger because it rules out the case when other controller's req match `vd.controller_owner_ref` but doesn't match etcd object's owners, so the delete will not happen and we are fine with that. But this case is complex to reason and I don't think there will be controller meeting this exact condition. Do you think I should just use the filter you wrote and ignore this corner case?

codyjrivera (approved):

src/v2/controllers/vdeployment_controller/trusted/rely_guarantee.rs
```diff
@@ -0,0 +1,190 @@
...
+pub open spec fn vd_rely_get_then_delete_req(req: GetThenDeleteRequest) -> StatePred<ClusterState> {
+    |s: ClusterState| {
+        req.key.kind == VReplicaSetView::kind() ==>
+            // preconditions not supported yet in GetThenDeleteRequest
+            !{
+                // req.owner_ref.controller.is_Some_and(true) && the same as vd.controller_owner_ref()
+                &&& exists |vd: VDeploymentView| req.owner_ref == vd.controller_owner_ref()
+                // is one of the owners of vrs
+                &&& s.resources().contains_key(req.key)
+                &&& s.resources()[req.key].metadata.owner_references_contains(req.owner_ref)
```
marshtompsxd, line 119: @Catoverflow I want to keep `vd_rely_get_then_delete_req` and `vd_rely_get_then_update_req` simple that they don't talk about `s.resources()`. Since `dyn` is not supported, I think the best solution now is to have a type invariant for `GetThenDeleteRequest` and `GetThenUpdateRequest` saying that their owner_ref has a true controller field. Then in `vd_rely_get_then_delete_req` and `vd_rely_get_then_update_req` you simply say that req.owner_ref's kind field cannot be a VD.

marshtompsxd: Essentially, I want to constrain the usage of `GetThenDeleteRequest` and `GetThenUpdateRequest`by type invariant (before we have `dyn` support) so that they can only carry a controller owner ref.

Catoverflow: @marshtompsxd 
> I want to keep vd_rely_get_then_delete_req and vd_rely_get_then_update_req simple that they don't talk about s.resources()

I think then we should have some invariants like `every_xxx_req_is_well_formed_and_etcd_obj_exists`. Is that what you are thinking of?

> have a type invariant for GetThenDeleteRequest and GetThenUpdateRequest saying that their owner_ref has a true controller field. 

How do you like https://github.com/anvil-verifier/anvil/pull/630/commits/fa74e818ff59ab84f69af4dda308ae0804249665

marshtompsxd: By type invariant, I am talking about https://github.com/verus-lang/verus/pull/1207

Let me know if this is not well supported by Verus

Catoverflow: Added note on type invariant of we are not using this now

PR #631, opened 2025-05-15
Move shared spec functions to `trusted`

Catoverflow:


src/v2/controllers/vdeployment_controller/trusted/util.rs
```diff
@@ -0,0 +1,100 @@
...
+    (new_vrs_list, old_vrs_list)
+}
+
+// See https://github.com/kubernetes/kubernetes/blob/cdc807a9e849b651fb48c962cc18e25d39ec5edf/pkg/controller/deployment/sync.go#L196-L210
+// pod template hash is used to prevent old and new vrs from owning the same pod
+// here we use resource_version of vd as a hash
+//
+// TODO: now we scale up the new vrs' replicas at once,
+// we may consider existing pods in old vrs later to satisfy maxSurge
+pub open spec fn make_replica_set(vd: VDeploymentView) -> (vrs: VReplicaSetView)
```
marshtompsxd, line 50: Is `make_replica_set` used in the liveness theorem? If not, it should not stay in trusted.

marshtompsxd: Same questions for other functions in this file. The trusted module should not import the proof (i.e., the model module) and still stay minimized

Catoverflow: > Is make_replica_set used in the liveness theorem? If not, it should not stay in trusted.

No, let me move it

PR #633, opened 2025-05-17
Use macro to define VDeployment

Catoverflow:
- Use macros to reduce LOC in CR
- Move `string_equal` to vstd_ext::string_view

### DIFF
--- a/src/v2/controllers/vdeployment_controller/trusted/exec_types.rs
+++ b/src/v2/controllers/vdeployment_controller/trusted/exec_types.rs
@@ -6,19 +6,11 @@ use crate::kubernetes_api_objects::exec::{
 };
 use crate::kubernetes_api_objects::spec::resource::*;
 use crate::vdeployment_controller::trusted::spec_types;
+use crate::vstd_ext::string_view::*;
 use deps_hack::kube::Resource;
 use vstd::prelude::*;
 
 verus! {
-
-// helper function to circumvent the lack of support for String in spec
-#[verifier(external_body)]
-pub fn string_equal(s1: &String, s2: &str) -> (res: bool)
-    ensures
-        s1@ == s2@ <==> res,
-{
-    s1 == s2
-}
 
 implement_custom_object_wrapper_type!(
     VDeployment,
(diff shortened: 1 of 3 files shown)


Catoverflow:
@marshtompsxd The followup discussion for https://github.com/anvil-verifier/anvil/pull/611#discussion_r2087171688

Should we have 2 different version as `well_formed` and `well_formed_and_namespaced`?

Another question is should we have setter in `ResourceView` trait? For example, `with_metadata`, `with_spec`, etc.

marshtompsxd:
> The followup discussion for https://github.com/anvil-verifier/anvil/pull/611#discussion_r2087171688
Should we have 2 different version as well_formed and well_formed_and_namespaced?

marshtompsxd:
> The followup discussion for https://github.com/anvil-verifier/anvil/pull/611#discussion_r2087171688
Should we have 2 different version as well_formed and well_formed_and_namespaced?

Just have one `well_formed_for_namespaced`

> Should we have setter in ResourceView trait? For example, with_metadata, with_spec, etc. I vote for this idea because we need it in implement_object_wrapper_type!, and should be called for all resource object wrapper.

You can add the setter in another PR. But why does the macro need the setter of the `ResourceView` trait?

> Should we have corresponding macros in spec/resource.rs? I can push another PR to do it.

By corresponding macros, do you mean macros for spec-level objects? I'm fine with it but please do it in a separate PR.

Catoverflow:
> But why does the macro need the setter of the ResourceView trait?

Because it's depended by macro in https://github.com/anvil-verifier/anvil/blob/c991951783082bcaa95766c3cc5c2e4b41bf5b9e/src/kubernetes_api_objects/exec/resource.rs#L41-L46  Or we have to add `with_metadata` when we use the macro like the change in `src/v2/controllers/vdeployment_controller/trusted/spec_types.rs` in this PR

Catoverflow:
> Just have one `well_formed_for_namespaced`

Let me put this in another PR. I think the current one is good to merge if there's nothing left to be done

src/v2/controllers/vdeployment_controller/trusted/exec_types.rs
```diff
@@ -20,39 +20,25 @@ pub fn string_equal(s1: &String, s2: &str) -> (res: bool)
     s1 == s2
 }
```
marshtompsxd, line 21: Minor: could you move this to string_view.rs?

marshtompsxd:
Could you improve the PR description a bit instead of having two bullets and one is crossed.

PR #634, opened 2025-05-17
Use macros in VRS controller

Catoverflow:


Catoverflow:
The error in CI is exactly what I had in #611 . @marshtompsxd Do you have some ideas on that?

Catoverflow:
Close as moved to #636

PR #637, opened 2025-05-21
Update `well_formed` to `well_formed_for_namespaced` and fix bug in exec

Catoverflow:
- Update `well_formed` to `well_formed_for_namespaced`
- Fix bug of namespace is checked in spec while not checked in exec

### DIFF
--- a/src/kubernetes_api_objects/exec/object_meta.rs
+++ b/src/kubernetes_api_objects/exec/object_meta.rs
@@ -187,12 +187,13 @@ impl ObjectMeta {
     }
 
     #[verifier(external_body)]
-    pub fn well_formed(&self) -> (b: bool)
-        ensures b == self@.well_formed(),
+    pub fn well_formed_for_namespaced(&self) -> (b: bool)
+        ensures b == self@.well_formed_for_namespaced(),
     {
         self.inner.name.is_some()
-            && self.inner.resource_version.is_some()
-            && self.inner.uid.is_some()
+        && self.inner.namespace.is_some()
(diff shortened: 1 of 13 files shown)


src/kubernetes_api_objects/exec/object_meta.rs
```diff
@@ -187,12 +187,13 @@ impl ObjectMeta {
     }
 
     #[verifier(external_body)]
```
marshtompsxd, line 189: @Catoverflow can you remove `#[verifier(external_body)]`? Your implementation should use the getter methods and the verification should pass.

Catoverflow: Sure, I forgot to do that. Should be all fixed now

Catoverflow:
But as `src/v2/controllers/vreplicaset_controller/trusted/spec_types.rs` is changed @codyjrivera may have to merge conflict if this PR goes into main first

src/kubernetes_api_objects/exec/object_meta.rs
```diff
@@ -49,6 +49,15 @@ impl ObjectMeta {
         self.inner.name.clone()
     }
 
+    #[verifier(external_body)]
+    pub fn uid(&self) -> (uid: Option<String>)
```
marshtompsxd, line 53: You don't need to add this. There are already `has_some_uid` and `has_some_resource_version` that you can use.

Catoverflow: fixed

PR #640, opened 2025-05-22
Make the HostId for Controller Messages Carry the Current Reconcile Key

codyjrivera:
I modify the `HostId` enum so that the `Controller` variant also carries an `ObjectRef`, indicating which reconcile key is associated with that message. I additionally modify the cluster state machine to handle this additional information, update definitions, and mark broken proofs for repair.

This new addition allows us to specify an invariant not only preventing reconciles from other installed controllers in the cluster from interfering with our reconcile, but also preventing other reconciles _from the controller we are reasoning about_ from interfering with our reconcile.

### DIFF
--- a/src/v2/kubernetes_cluster/spec/cluster.rs
+++ b/src/v2/kubernetes_cluster/spec/cluster.rs
@@ -278,7 +278,8 @@ impl Cluster {
         Action {
             precondition: |input: (Option<Message>, Option<ObjectRef>), s: ClusterState| {
                 &&& self.controller_models.contains_key(controller_id)
-                &&& received_msg_destined_for(input.0, HostId::Controller(controller_id))
+                &&& input.1.is_Some()
+                &&& received_msg_destined_for(input.0, HostId::Controller(controller_id, input.1.get_Some_0()))
                 &&& result(input, s).0.is_Enabled()
                 &&& result(input, s).1.is_Enabled()
             },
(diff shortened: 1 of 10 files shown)


Catoverflow (approved):

src/v2/kubernetes_cluster/spec/cluster.rs
```diff
@@ -278,7 +278,8 @@ impl Cluster {
         Action {
             precondition: |input: (Option<Message>, Option<ObjectRef>), s: ClusterState| {
                 &&& self.controller_models.contains_key(controller_id)
-                &&& received_msg_destined_for(input.0, HostId::Controller(controller_id))
+                &&& input.1.is_Some()
+                &&& received_msg_destined_for(input.0, HostId::Controller(controller_id, input.1.get_Some_0()))
```
marshtompsxd, line 282: Is it true that input.1 should ALWAYS be Some here (for the state machine to take next action)?

src/v2/kubernetes_cluster/spec/cluster.rs
```diff
@@ -679,7 +680,8 @@ impl Cluster {
             let network_result = network().next_result(msg_ops, s.network);
 
             &&& self.controller_models.contains_key(input.0)
-            &&& received_msg_destined_for(input.1, HostId::Controller(controller_id))
+            &&& input.1.is_Some()
+            &&& received_msg_destined_for(input.1, HostId::Controller(controller_id, input.2.get_Some_0()))
```
marshtompsxd, line 684: Same question as above

PR #641, opened 2025-05-30
Update Verus Version to 3b6b805ac

Catoverflow:
- Update Verus Version to 3b6b805ac
- Mask buggy lifetime check of Verus before https://github.com/verus-lang/verus/issues/1708 is fixed

This PR is the first stage of catching up with Verus' rolling release. We will work with @Chris-Hawblitzel to add Anvil to veritas as both test and performance benchmark. We need further discussions on stabilizing a release branch for this usage.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -17,4 +17,5 @@ cd ..
 "$rv" -L dependency=deps_hack/target/debug/deps \
   --extern=deps_hack="deps_hack/target/debug/libdeps_hack.rlib" \
   --compile \
+  --no-lifetime \
   "$@"
--- a/src/executable_model/common.rs
+++ b/src/executable_model/common.rs
@@ -589,13 +589,13 @@ impl ResourceView for SimpleCRView {
 
     proof fn marshal_preserves_kind() {}
 
-    closed spec fn marshal_spec(s: SimpleCRSpecView) -> Value;
+    uninterp spec fn marshal_spec(s: SimpleCRSpecView) -> Value;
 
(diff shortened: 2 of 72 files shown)


src/executable_model/common.rs
```diff
@@ -373,6 +373,7 @@ impl Role {
                 invariant
                     all_valid == (forall |j| #![trigger policy_rules[j]] 0 <= j < i ==> policy_rules@.map_values(|policy_rule: PolicyRule| policy_rule@)[j].state_validation()),
                     i <= policy_rules.len(),
+                decreases policy_rules.len() - i,
```
codyjrivera, line 376: Did verus previously support inferring ranking functions? If so, this makes it a regression.

Catoverflow: > Did verus previously support inferring ranking functions?

They didn't check, just enabled in a recent PR: https://github.com/verus-lang/verus/pull/1545

Chris-Hawblitzel: You may also be able to do `for i in 0..policy_rules.len()` instead of `while`, and for `for` the `decreases` is inferred automatically.

build.sh
```diff
@@ -17,4 +17,5 @@ cd ..
 "$rv" -L dependency=deps_hack/target/debug/deps \
   --extern=deps_hack="deps_hack/target/debug/libdeps_hack.rlib" \
   --compile \
+  --no-lifetime \
```
marshtompsxd, line 20: @Catoverflow can you add a TODO here? We need to revisit this lifetime thing.

PR #643, opened 2025-06-04
Fix off-by-one error in controller action precondition

codyjrivera:
This fixes an off-by-one error in `controller_action_pre`: instead of forcing the `ObjectRef` to be non-null, it forces the message to be `null`, whereas it is allowed that a controller takes a step without a message as input.

This is separate from the rest of subtask in (https://github.com/xlab-uiuc/anvil/issues/26) since spec/trusted code is modified.

### DIFF
--- a/src/v2/kubernetes_cluster/spec/cluster.rs
+++ b/src/v2/kubernetes_cluster/spec/cluster.rs
@@ -680,7 +680,7 @@ impl Cluster {
             let network_result = network().next_result(msg_ops, s.network);
 
             &&& self.controller_models.contains_key(input.0)
-            &&& input.1.is_Some()
+            &&& input.2.is_Some()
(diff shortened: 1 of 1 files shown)


marshtompsxd:
Good catch. I'm surprised that this bug didn't break many proofs before. How did you discover it?

codyjrivera:
Fixing the WF1 proof `lemma_from_init_step_to_send_list_pods_req`. It was already broken by #640, and I marked it as such.

Catoverflow:
I'm also curious why this didn't break proofs as a non-null `ObjectRef` is required
https://github.com/anvil-verifier/anvil/blob/c10a78a71368c3307d3d692c2359a048ec083e67/src/v2/kubernetes_cluster/spec/cluster.rs#L279-L285

Catoverflow (approved):

marshtompsxd (approved):

PR #644, opened 2025-06-10
Prove VDeployment controller eventually terminates

Catoverflow:
- Prove termination for VDeployment controller in `src/v2/controllers/vdeployment_controller/proof/liveness/terminate.rs`
- Add trait `IntoSpec` for easier encoding of step predicates. This may not be implemented in the best way but it works!

(Good) Side effects:
- make `tla_exists_proved_by_witness` public because it's used in termination proof
- lift trigger to TLA embedding for one VRS proof
- move some utility functions into from `src/v2/controllers/vdeployment_controller/model/reconciler.rs` `src/v2/controllers/vdeployment_controller/trusted/util.rs`. So we don't need another version of spec function emulating the actual logic of controller in high level predicate (`current_state_matches`)
- fix bug in `current_state_matches`

### DIFF
--- a/src/temporal_logic/rules.rs
+++ b/src/temporal_logic/rules.rs
@@ -70,6 +70,7 @@ proof fn tla_exists_unfold<T, A>(ex: Execution<T>, a_to_p: spec_fn(A) -> TempPre
     ensures exists |a| #[trigger] a_to_p(a).satisfied_by(ex),
 {}
 
+// TODO: add a lemma to prove by witness and elimiate the reasoning over execution
 pub proof fn tla_exists_proved_by_witness<T, A>(ex: Execution<T>, a_to_p: spec_fn(A) -> TempPred<T>, witness_a: A)
     requires a_to_p(witness_a).satisfied_by(ex),
     ensures tla_exists(a_to_p).satisfied_by(ex)


src/v2/controllers/vreplicaset_controller/proof/liveness/terminate.rs
```diff
@@ -316,7 +316,7 @@ pub proof fn reconcile_eventually_terminates_on_vrs_object(
     assert forall |n: nat| #![auto]
         spec.entails(lift_state(at_step_state_pred(controller_id, vrs, VReplicaSetRecStepView::AfterCreatePod(n)))
                     .leads_to(lift_state(reconcile_idle))) by {
-        assert forall |n: nat| #![trigger after_create_pod_rank(controller_id, vrs, n)]
+        assert forall |n: nat| #![trigger lift_state(after_create_pod_rank(controller_id, vrs, n))]
```
marshtompsxd, line 319: Why change the trigger to the lift version?

src/kubernetes_api_objects/exec/object_meta.rs
```diff
@@ -49,6 +49,15 @@ impl ObjectMeta {
         self.inner.name.clone()
     }
 
+    #[verifier(external_body)]
```
marshtompsxd: @Catoverflow Do you really need this function?

src/v2/controllers/vreplicaset_controller/proof/liveness/terminate.rs
Catoverflow: I was testing on which layer the trigger could work so I can put it in VD proof. As trigger on TLA embedding works it's better than the original one so I didn't change back.

src/temporal_logic/rules.rs
```diff
@@ -70,7 +70,7 @@ proof fn tla_exists_unfold<T, A>(ex: Execution<T>, a_to_p: spec_fn(A) -> TempPre
     ensures exists |a| #[trigger] a_to_p(a).satisfied_by(ex),
 {}
 
-proof fn tla_exists_proved_by_witness<T, A>(ex: Execution<T>, a_to_p: spec_fn(A) -> TempPred<T>, witness_a: A)
+pub proof fn tla_exists_proved_by_witness<T, A>(ex: Execution<T>, a_to_p: spec_fn(A) -> TempPred<T>, witness_a: A)
```
marshtompsxd, line 74: Do you really need to use this lemma? (I ask because its proof body is empty)

src/v2/controllers/vdeployment_controller/proof/predicate.rs
```diff
@@ -0,0 +1,396 @@
...
+        ensures
+            #[trigger] self.into_temporal_pred(controller_id, vd) == lift_state(self.into_cluster_state_pred(controller_id, vd)),
+            #[trigger] self.into_temporal_pred(controller_id, vd) == lift_state(
+                Cluster::at_expected_reconcile_states(controller_id, vd.object_ref(), self.into_local_state_pred())
+            )
+    {
+        self.lemma_cluster_equiv(controller_id, vd);
+    }
+}
+
```
marshtompsxd, line 102: @Catoverflow could you leave a TODO saying that we will explore a better way to shortcut disjunction/conjunction of state predicates?

PR #645, opened 2025-06-10
Use macro instead of trait in VD termination proof

Catoverflow:
Follow up of #644
Managed to use macro instead of trait, made several compromising:
- Use dummy trigger
- Use wrapper of macro as trigger
- Use trigger at inner layer
But again it works and proof isn't slow


Catoverflow:
Closed as it's chaotic to rebase, will use cherry-pick to fix this

PR #646, opened 2025-06-10
Use macro for VD termination proof

Catoverflow:
Follow up of https://github.com/anvil-verifier/anvil/pull/644
Managed to use macro instead of trait, made several compromising:

- Use dummy trigger
- Use wrapper of macro as trigger
- Use trigger at inner layer
- But again it works and proof isn't slow

### DIFF
--- a/src/v2/controllers/vdeployment_controller/proof/predicate.rs
+++ b/src/v2/controllers/vdeployment_controller/proof/predicate.rs
@@ -14,383 +14,56 @@ use vstd::prelude::*;
 
 verus! {
 
-// Represents a reconcile step (optionally with a predicate) for use in spec functions.
-//
-// Usage:
-//
-// Step-only case
-// let step1 = StepCase::Plain(Step1);
-//
-// Step with an additional predicate over VDeploymentReconcileState
-// let step2 = StepCase::WithPred(Step2, |vds: VDeploymentReconcileState| vds.some_field == 5);
-//
-// Convert into a predicate over ReconcileLocalState (used in Cluster::at_expected_reconcile_states)
-// let pred_local: spec_fn(ReconcileLocalState) -> bool = step1.into_local_state_pred();
-//
-// Convert into a predicate over ClusterState
-// let pred_cluster: spec_fn(ClusterState) -> bool = step2.into_cluster_state_pred(controller_id, vd);
-//
-// Convert into a temporal predicate over ClusterState (for use in TempPred context)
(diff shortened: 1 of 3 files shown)


src/v2/controllers/vdeployment_controller/proof/predicate.rs
```diff
@@ -14,383 +14,56 @@ use vstd::prelude::*;
...
-                Cluster::at_expected_reconcile_states(controller_id, vd.object_ref(), self.into_local_state_pred())
-            )
-    {
-        self.lemma_cluster_equiv(controller_id, vd);
-    }
+// usage: at_step![step,*]
+// step_or_pred = step | (step, pred)
+#[macro_export]
+macro_rules! at_step_or {
+    [ $($tokens:tt)+ ] => {
```
marshtompsxd, line 37: Why use `[]` instead of `()`?

src/v2/controllers/vdeployment_controller/trusted/step.rs
```diff
@@ -46,6 +46,7 @@ impl View for VDeploymentReconcileStep {
 }
 
 #[is_variant]
+#[derive(PartialEq)]
```
marshtompsxd: Do you need this trait to do the proof?

src/v2/controllers/vdeployment_controller/proof/predicate.rs
Catoverflow: I think `[]` makes it more clear that these steps are commutable

src/v2/controllers/vdeployment_controller/trusted/step.rs
Catoverflow: No, removed

src/v2/controllers/vdeployment_controller/trusted/step.rs
```diff
@@ -56,4 +56,10 @@ pub enum VDeploymentReconcileStepView {
     Error,
 }
 
+impl VDeploymentReconcileStepView {
+    pub open spec fn eq_step(self, other: VDeploymentReconcileStepView) -> bool {
```
marshtompsxd, line 64: Could you just replace this eq_step with ==? After that you can merge the PR

Catoverflow: > Could you just replace this eq_step with ==? After that you can merge the PR

No, this is to babysit Verus, I left comment in `predicate.rs`, let me move it here

PR #647, opened 2025-06-16
Fix WF1 proofs on 'top' (liveness) side of Bridge Invariant

codyjrivera:
This PR rewrites the liveness proof to use a bridge invariant saying that no message potentially interfering with the current reconcile for `vrs` can be in the network, unless the source is precisely the reconcile for `vrs`. The fixes are complete minus a helper lemma, and lemmas specifying the results of the API server taking an action.

### DIFF
--- a/src/v2/controllers/vreplicaset_controller/proof/liveness/api_actions.rs
+++ b/src/v2/controllers/vreplicaset_controller/proof/liveness/api_actions.rs
@@ -244,6 +244,43 @@ pub proof fn lemma_create_matching_pod_request_adds_matching_pod_and_returns_ok(
     return handle_create_request_msg(cluster.installed_types, msg, s.api_server).1;
 }
 
+// TODO: Prove this (I imagine we'll need a subtly different precondition and postcondition).
+#[verifier(external_body)]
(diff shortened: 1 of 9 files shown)


src/v2/controllers/vreplicaset_controller/proof/liveness/api_actions.rs
```diff
@@ -18,6 +18,271 @@ use vstd::{map::*, map_lib::*, prelude::*};
...
+                s.resources()[filtered_pod_keys[diff as int]]
+            )
+        }),
+        // should be an obvious corollary of `generated_name_is_unique`.
+        matching_pods(vrs, s.resources()).len() == matching_pods(vrs, s_prime.resources()).len() + 1,
+{
+    return handle_get_then_delete_request_msg(msg, s.api_server).1;
+}
+
+// OBSOLETE FUNCTIONS ------ Remove as others get used.
```
codyjrivera, line 284: There are still invariant proofs relying on these, hence why I didn't delete them yet.

codyjrivera:
Yes; I plan further changes (proving unproven lemmas, etc) for  other PRs.

PR #648, opened 2025-06-19
Fix VDeployment reconciler bug

Catoverflow:
Fix a bug when there are multiple VReplicaset matching VDeployment's template, VDeployment controller can not reconcile the cluster state to be desired state by scaling down all but one of them.

### Bug description

Previously the check on managed ReplicaSets was

```rust
if new_vrs_list.len() == 0 {
    // create the new vrs
    create_new_vrs(state, vd)
} else {
    let new_vrs = new_vrs_list.last();
    let state = VDeploymentReconcileState {
        new_vrs: Some(new_vrs),
        ..state
    };
    // ..
```

Because when replicating the logic of k8s ReplicaSet(RS) controller, there was one [comment](https://github.com/kubernetes/kubernetes/blob/73014647f9bdd7186b5fe121326b961d6521c162/pkg/controller/deployment/util/deployment_util.go#L633-L636):
```rust
 // In rare cases, such as after cluster upgrades, Deployment may end up with 
 // having more than one new ReplicaSets that have the same template as its template, 
 // see https://github.com/kubernetes/kubernetes/issues/40415 
 // We deterministically choose the oldest new ReplicaSet.
```
In kubernetes, when there are multiple RS with the same template as a deployment's, only one will be marked as new(updated) replicaset. Back that time I don't think our controller will allow multiple vrs to co-exist under the same template, but that is possible under the model of pod monkey. In this case, the deployment controller will not be able to reconcile cluster state with multiple RS matching deployment CR's template to desired state with only one RS. It simply ignores other RS.

Vanilla kubernetes implementation doesn't have this problem because it uses negation to select all - new as old RS to be scaled down, no matter if their template match deployment's or not. However, my model only sort out RS with template unmatched to be scaled down. This case is also a corner case caused by updating kubernetes itself, and slipped through our e2e test but can be captured by the pod monkey model. It's not found by verification, it's found by specification when I was trying to find a correct way to describe the current state.

### Flaky proof

Due to the flakiness of Verus proof, `vdeployment_controller::exec::reconciler::reconcile_core` is masked, while it can be manually proved with
```shell
verus src/v2_vdeployment_controller.rs -L dependency=src/deps_hack/target/debug/deps --extern=deps_hack="src/deps_hack/target/debug/libdeps_hack.rlib" --verify-only-module vdeployment_controller::exec::reconciler --no-lifetime --verify-function reconcile_core
```
Detailed report can be found at https://github.com/verus-lang/verus/issues/1756.

### DIFF
--- a/src/v2/controllers/vdeployment_controller/exec/reconciler.rs
+++ b/src/v2/controllers/vdeployment_controller/exec/reconciler.rs
@@ -136,7 +136,7 @@ pub fn reconcile_core(vd: &VDeployment, resp_o: Option<Response<VoidEResp>>, sta
             let (new_vrs, old_vrs_list) = filter_old_and_new_vrs(vd, filter_vrs_list(vrs_list_or_none.clone().unwrap(), vd));
             // no .last().cloned() in verus because "The verifier does not yet support the following Rust feature: overloaded deref"
             let state = VDeploymentReconcileState {
-                new_vrs: new_vrs,
+                new_vrs: new_vrs.clone(),
                 old_vrs_list: old_vrs_list.clone(),
                 ..state
             };
--- a/src/v2/controllers/vdeployment_controller/exec/reconciler.rs
+++ b/src/v2/controllers/vdeployment_controller/exec/reconciler.rs
@@ -128,24 +128,18 @@ pub fn reconcile_core(vd: &VDeployment, resp_o: Option<Response<VoidEResp>>, sta
             if vrs_list_or_none.is_none() {
                 return (error_state(state), None);
             }
-            let (new_vrs_list, old_vrs_list) = filter_old_and_new_vrs(filter_vrs_list(vrs_list_or_none.clone().unwrap(), vd), vd);
+            let (new_vrs, old_vrs_list) = filter_old_and_new_vrs(vd, filter_vrs_list(vrs_list_or_none.clone().unwrap(), vd));
             // no .last().cloned() in verus because "The verifier does not yet support the following Rust feature: overloaded deref"
             let state = VDeploymentReconcileState {
-                reconcile_step: VDeploymentReconcileStep::Error,
-                new_vrs: None,
+                new_vrs: new_vrs,
                 old_vrs_list: old_vrs_list.clone(),
+                ..state
             };
-            if new_vrs_list.len() == 0 {
+            if new_vrs.is_none() {
                 // no new vrs, create one
                 return create_new_vrs(state, vd);
             }
-            // verus doesn't have .pop() equivalent, and .last().cloned() is not supported either
-            let new_vrs = new_vrs_list[new_vrs_list.len() - 1].clone();
-            let state = VDeploymentReconcileState {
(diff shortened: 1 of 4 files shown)


marshtompsxd:
@Catoverflow Can you write more details about this bug? What did the controller do wrong? What should be the correct behavior? How do you fix it?

Catoverflow:
Previously the check on managed ReplicaSets was
```rust
if new_vrs_list.len() == 0 {
    // create the new vrs
    create_new_vrs(state, vd)
} else {
    let new_vrs = new_vrs_list.last();
    let state = VDeploymentReconcileState {
        new_vrs: Some(new_vrs),
        ..state
    };
    // ..
```
Because when replicating the logic of k8s ReplicaSet(RS) controller, there was one comment:
```go
 // In rare cases, such as after cluster upgrades, Deployment may end up with 
 // having more than one new ReplicaSets that have the same template as its template, 
 // see https://github.com/kubernetes/kubernetes/issues/40415 
 // We deterministically choose the oldest new ReplicaSet.
```
In kubernetes, when there are multiple RS with the same template as a deployment's, only one will be marked as new(updated) replicaset. Back that time I don't think our controller will allow multiple vrs to co-exist under the same template, but that is possible under the model of pod monkey. In this case, the deployment controller will not be able to reconcile cluster state with multiple RS matching deployment CR's template to desired state with only one RS. It simply ignores other RS.

Vanilla kubernetes implementation doesn't have this problem because it uses negation to select all - new as old replicasets to be scaled down, no matter if their template match deployment's or not. However, my model only sort out replicasets with template unmatched to be scaled down. This case is also a corner case caused by updating kubernetes itself, and slipped through our e2e test.

src/v2/controllers/vdeployment_controller/exec/reconciler.rs
```diff
@@ -98,9 +98,19 @@ pub fn reconcile_error(state: &VDeploymentReconcileState) -> (res: bool)
 // 4. rollback is not supported.
 // *. When having multiple new vrs, k8s deterministically chooses the oldest one.
 //    We may not need to support this if we can prove this doesn't happen for our controller.
+ // mask this proof before there's a solution to flakiness
+ // see https://github.com/verus-lang/verus/issues/1756
+#[verifier(external_body)]
```
marshtompsxd: Does the hotfix address the proof failure?

Catoverflow: Yes. Without this mask it can be verified by
```shell
verus src/v2_vdeployment_controller.rs -L dependency=src/deps_hack/target/debug/deps --extern=deps_hack="src/deps_hack/target/debug/libdeps_hack.rlib" --verify-only-module vdeployment_controller::exec::reconciler --no-lifetime --verify-function reconcile_core --expand-errors
```
While cannot pass `build.sh` which verifiers all the other proofs. Probably due to the flakiness I guess.

marshtompsxd:
@Catoverflow Can you address the two remaining CI failures and update the PR desc with the detailed bug report as well as this weird verification failure and how we workaround it?

PR #651, opened 2025-06-23
Change rely-guarantee properties for transactional API

codyjrivera:
Here I update the rely and guarantee conditions for `VReplicaSet` dealing with messages using the transactional API to not explicitly discuss the state of etcd. Instead, they merely put conditions on the kind of the owner reference sent along with the request. The API server will ensure that this owner reference and version number of the object in etcd matches, aborting otherwise.

These changes are analogous to those already in `VDeployment`. They break a couple of proofs I was already expecting to delete soon anyway.

### DIFF
--- a/src/controllers/vreplicaset_controller/trusted/rely_guarantee.rs
+++ b/src/controllers/vreplicaset_controller/trusted/rely_guarantee.rs
@@ -48,6 +48,30 @@ pub open spec fn vrs_rely_update_req(req: UpdateRequest) -> StatePred<ClusterSta
     }
 }
 
+// Update requests to pods must carry a resource version on them
+// and cannot 1) update pods owned by a VReplicaSet
+//         or 2) update pod to become owned
+// by a VReplicaSet.
+pub open spec fn vrs_rely_get_then_update_req(req: GetThenUpdateRequest) -> StatePred<ClusterState> {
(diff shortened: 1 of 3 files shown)


src/controllers/vreplicaset_controller/trusted/rely_guarantee.rs
```diff
@@ -48,6 +48,30 @@ pub open spec fn vrs_rely_update_req(req: UpdateRequest) -> StatePred<ClusterSta
...
+    |s: ClusterState| {
+        req.obj.kind == Kind::PodKind ==> {
+            // Prevents 1): where other controllers update pods owned by a VReplicaSet.
+            // an object can has multiple owners, but only one controller owner
+            // We force requests from other controllers to carry the controller owner reference
+            // to achieve exclusive ownerships
+            // TODO: add type invariant
+            &&& req.owner_ref.controller.is_Some()
+            &&& req.owner_ref.controller.get_Some_0()
+            &&& req.owner_ref.kind != Kind::PodKind
```
marshtompsxd: This should be `req.owner_ref.kind != VRS`, right?

codyjrivera: oops, thanks for catching.

PR #654, opened 2025-06-25
Update VD reconciler util functions

Catoverflow:
Update VD reconciler utility functions' signature to make it easier to reason on input for liveness proofs.
Add deepview trait implementation for `DynamicObject`

### DIFF
--- a/src/controllers/vdeployment_controller/exec/reconciler.rs
+++ b/src/controllers/vdeployment_controller/exec/reconciler.rs
@@ -320,11 +320,11 @@ ensures
     return (state_prime, Some(Request::KRequest(req)))
 }
 
-fn match_replicas(vrs: &VReplicaSet, vd: &VDeployment) -> (res: bool)
+fn match_replicas(vd: &VDeployment, vrs: &VReplicaSet) -> (res: bool)
 requires
     vd@.well_formed(),
     vrs@.well_formed(),
-ensures res == model_reconciler::match_replicas(vrs@, vd@),
+ensures res == model_reconciler::match_replicas(vd@, vrs@),
 {
     vrs.spec().replicas().unwrap_or(1) == vd.spec().replicas().unwrap_or(1)
 }
--- a/src/controllers/vdeployment_controller/exec/reconciler.rs
+++ b/src/controllers/vdeployment_controller/exec/reconciler.rs
@@ -29,7 +29,7 @@ impl View for VDeploymentReconcileState {
         model_reconciler::VDeploymentReconcileState {
             reconcile_step: self.reconcile_step@,
             new_vrs: self.new_vrs.deep_view(),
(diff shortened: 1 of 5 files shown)


src/controllers/vdeployment_controller/exec/reconciler.rs
```diff
@@ -236,13 +239,12 @@ pub fn done_state(state: VDeploymentReconcileState) -> (state_prime: VDeployment
...
 // create new vrs
-pub fn create_new_vrs(state: VDeploymentReconcileState, vd: &VDeployment) -> (res: (VDeploymentReconcileState, Option<Request<VoidEReq>>))
+pub fn create_new_vrs(old_vrs_list: Vec<VReplicaSet>, vd: &VDeployment) -> (res: (VDeploymentReconcileState, Option<Request<VoidEReq>>))
 requires
     vd@.well_formed(),
-    state.new_vrs@.is_None(),
 ensures
-    res.1@.is_Some() && model_reconciler::create_new_vrs(state@, vd@).1.is_Some(),
-    (res.0@, res.1.deep_view()) == model_reconciler::create_new_vrs(state@, vd@),
+    res.1@.is_Some() && model_reconciler::create_new_vrs(old_vrs_list.deep_view(), vd@).1.is_Some(),
```
marshtompsxd: @Catoverflow Why do you need a postcondition asserting that the returned request is always Some?

Catoverflow: @marshtompsxd Because this `create_new_vrs` is guaranteed to return a request

marshtompsxd: I understand. But why do we need this postcondition? All we need is to say that the exec behaviors like the model.

Catoverflow: This wrapper is a part of exec code? Though it's already removed in the most recent commits.

src/controllers/vdeployment_controller/exec/reconciler.rs
```diff
@@ -100,6 +100,7 @@ pub fn reconcile_error(state: &VDeploymentReconcileState) -> (res: bool)
 //    We may not need to support this if we can prove this doesn't happen for our controller.
  // mask this proof before there's a solution to flakiness
  // see https://github.com/verus-lang/verus/issues/1756
+#[verifier(external_body)]
```
marshtompsxd: @Catoverflow So the proof is still broken?

src/controllers/vdeployment_controller/exec/reconciler.rs
marshtompsxd: oh I misread

marshtompsxd:
@Catoverflow Could you briefly explain why you inline a lot of functions?

PR #657, opened 2025-06-30
Add top-level proof harness for VDeployment controller

codyjrivera:
Without expounding on details/invariants specific to the controller; I provide a top-level proof wrapper for `VDeployment`.

Unfortunately, since `VDeployment` imports `VReplicaSet`, proving `VDeployment` also proves `VReplicaSet`, and a proof `lemma_current_state_matches_is_stable` has become flaky (when proved alongside `VDeployment`) as a result.


marshtompsxd:
@codyjrivera Why not just use `--verify-module` in the CI to focus on the VD controller-specific proofs? In that way, you should not need to edit anything in the VRS controller's proof.

PR #662, opened 2025-07-06
WF1 proof for VDeployment controller

Catoverflow:
- Patch ESR to use predicates on etcd state
- Complete WF1 top level proof for VDeployment controller
  - mask broken transition lemmas for future PR

Minor change:
- unify `dummy` trigger for easier usage
- as `vreplicaset_controller::trusted` is the only module required by VDeployment controller, update the `pub_mod.rs`

### DIFF
--- a/src/controllers/vdeployment_controller/proof/liveness/resource_match.rs
+++ b/src/controllers/vdeployment_controller/proof/liveness/resource_match.rs
@@ -119,183 +119,167 @@ ensures
             }
             temp_pred_equality(list_resp_msg(msg), tla_exists(|i: (Option<int>, nat)| after_list_with_etcd_state(msg, i.0, i.1)));
         }
-        // \A |n| \A |replicas| etcd_state_is(replicas, n) ~> after_ensure_vrs(n)
-        assert forall |n: nat| #![trigger after_ensure_vrs(n)] spec.entails(tla_exists(|replicas: Option<int>| after_list_with_etcd_state(msg, replicas, n))
-            .leads_to(after_ensure_vrs(n))) by {
-            // \A |replicas| after_list_with_etcd_state(msg, Some(replicas), n) ~> after_ensure_vrs(n)
-            // since here the transitions branch over the existence and replicas of new vrs
-            assert forall |replicas: Option<int>| spec.entails(#[trigger] after_list_with_etcd_state(msg, replicas, n).leads_to(after_ensure_vrs(n))) by {
-                // new vrs does not exists. Here the existance is encoded as is_Some, and replicas is get_Some_0
-                if replicas.is_None() {
-                    // AfterListVRS ~> AfterCreateNewVRS
-                    let create_vrs_req = lift_state(and!(
-                        at_vd_step_with_vd(vd, controller_id, at_step![(AfterEnsureNewVRS, local_state_is(Some(vd.spec.replicas.unwrap_or(int1!())), n))]),
-                        pending_create_new_vrs_req_in_flight(vd, controller_id),
-                        etcd_state_is(vd, controller_id, None, n),
-                        local_state_match_etcd(vd, controller_id)
-                    ));
-                    assert(spec.entails(after_list_with_etcd_state(msg, None, n).leads_to(create_vrs_req))) by {
-                        lemma_from_after_receive_list_vrs_resp_to_send_create_new_vrs_req(vd, spec, cluster, controller_id, msg, n);
+        // \A |replicas, n| etcd_state_is(replicas, n) ~> \E |n| after_ensure_vrs(n)
+        assert forall |i: (Option<int>, nat)| #![trigger after_list_with_etcd_state(msg, i.0, i.1)] spec.entails(after_list_with_etcd_state(msg, i.0, i.1).leads_to(tla_exists(|n| after_ensure_vrs(n)))) by {
+            let (replicas, n) = i;
+            // new vrs does not exists. Here the existance is encoded as is_Some, and replicas is get_Some_0
+            if replicas.is_None() {
+                // AfterListVRS ~> AfterCreateNewVRS
+                let create_vrs_req = lift_state(and!(
+                    at_vd_step_with_vd(vd, controller_id, at_step![(AfterEnsureNewVRS, local_state_is(Some(vd.spec.replicas.unwrap_or(int1!())), n))]),
+                    pending_create_new_vrs_req_in_flight(vd, controller_id),
+                    etcd_state_is(vd, controller_id, None, n),
+                    local_state_match_etcd(vd, controller_id)
+                ));
+                assert(spec.entails(after_list_with_etcd_state(msg, None, n).leads_to(create_vrs_req))) by {
+                    lemma_from_after_receive_list_vrs_resp_to_send_create_new_vrs_req(vd, spec, cluster, controller_id, msg, n);
+                }
+                let create_vrs_req_msg = |msg: Message| lift_state(and!(
+                    at_vd_step_with_vd(vd, controller_id, at_step![(AfterEnsureNewVRS, local_state_is(Some(vd.spec.replicas.unwrap_or(int1!())), n))]),
+                    req_msg_is_pending_create_new_vrs_req_in_flight(vd, controller_id, msg),
+                    etcd_state_is(vd, controller_id, None, n),
+                    local_state_match_etcd(vd, controller_id)
+                ));
+                assert(create_vrs_req == tla_exists(|msg| create_vrs_req_msg(msg))) by {
+                    assert forall |ex: Execution<ClusterState>| #[trigger] create_vrs_req.satisfied_by(ex) implies
+                        tla_exists(|msg| create_vrs_req_msg(msg)).satisfied_by(ex) by {
+                        let s = ex.head();
+                        let req_msg = s.ongoing_reconciles(controller_id)[vd.object_ref()].pending_req_msg.get_Some_0();
+                        assert((|msg| create_vrs_req_msg(msg))(req_msg).satisfied_by(ex));
                     }
-                    let create_vrs_req_msg = |msg: Message| lift_state(and!(
-                        at_vd_step_with_vd(vd, controller_id, at_step![(AfterEnsureNewVRS, local_state_is(Some(vd.spec.replicas.unwrap_or(int1!())), n))]),
-                        req_msg_is_pending_create_new_vrs_req_in_flight(vd, controller_id, msg),
-                        etcd_state_is(vd, controller_id, None, n),
+                    temp_pred_equality(create_vrs_req, tla_exists(|msg| create_vrs_req_msg(msg)));
+                }
+                let create_vrs_resp = lift_state(and!(
+                    at_vd_step_with_vd(vd, controller_id, at_step![(AfterCreateNewVRS, local_state_is(Some(vd.spec.replicas.unwrap_or(int1!())), n))]),
+                    exists_resp_msg_is_ok_create_new_vrs_resp(vd, controller_id),
+                    etcd_state_is(vd, controller_id, Some(vd.spec.replicas.unwrap_or(int1!())), n),
+                    local_state_match_etcd(vd, controller_id)
+                ));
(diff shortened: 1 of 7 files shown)


src/temporal_logic/rules.rs
```diff
@@ -682,6 +682,12 @@ pub proof fn leads_to_exists_intro<T, A>(spec: TempPred<T>, a_to_p: spec_fn(A) -
     };
 }
 
+#[verifier(external_body)]
```
marshtompsxd: Why mask it?

src/controllers/vdeployment_controller/trusted/liveness_theorem.rs
```diff
@@ -20,20 +20,36 @@ pub open spec fn vd_eventually_stable_reconciliation_per_cr(vd: VDeploymentView)
...
+    let objs = resources.values().filter(|obj: DynamicObjectView| {
+        &&& obj.kind == VReplicaSetView::kind()
+        &&& obj.metadata.namespace == vd.metadata.namespace
+    }).to_seq();
+    let vrs_list = objects_to_vrs_list(objs);
+    let filtered_vrs_list = filter_vrs_list(vd, vrs_list.unwrap());
+    filter_old_and_new_vrs(vd, filtered_vrs_list)
+}
+
+// why this fn makes proof pass?
```
marshtompsxd: Don't just put a question in the comment. Make it either a TODO for future investigation, or explain why things work.

src/controllers/vdeployment_controller/trusted/liveness_theorem.rs
```diff
@@ -20,20 +20,36 @@ pub open spec fn vd_eventually_stable_reconciliation_per_cr(vd: VDeploymentView)
...
     |s: ClusterState| {
-        let objs = s.resources().values().filter(|obj: DynamicObjectView| {
-            &&& obj.kind == VReplicaSetView::kind()
-            &&& obj.metadata.namespace == vd.metadata.namespace
-        }).to_seq();
-        let vrs_list = objects_to_vrs_list(objs);
-        let filtered_vrs_list = filter_vrs_list(vd, vrs_list.unwrap());
-        let (new_vrs, old_vrs_list) = filter_old_and_new_vrs(vd, filtered_vrs_list);
-        &&& vrs_list.is_Some()
+        // simulate handle_list_req
```
marshtompsxd: Is this change for making the spec consistent with API server's state machine so the proof will be easier?

src/controllers/vdeployment_controller/proof/liveness/resource_match.rs
```diff
@@ -0,0 +1,1516 @@
...
+        &&& spec.entails(tla_forall(|i: (Option<Message>, Option<ObjectRef>)| cluster.controller_next().weak_fairness((controller_id, i.0, i.1))))
+    };
+    // Init ~> send list req
+    let init = lift_state(and!(at_vd_step_with_vd(vd, controller_id, at_step![Init]), no_pending_req_in_cluster(vd, controller_id)));
+    let list_req = lift_state(and!(at_vd_step_with_vd(vd, controller_id, at_step![AfterListVRS]), pending_list_req_in_flight(vd, controller_id)));
+    assert(spec.entails(init.leads_to(list_req))) by {
+        lemma_from_init_step_to_send_list_vrs_req(vd, spec, cluster, controller_id);
+    }
+    // send list req ~> exists |msg| list_req_msg(msg)
+    // TODO: try to use X_req == tla_exists(|msg| X_req_msg(msg)) in the future
```
marshtompsxd: Seems that this TODO is already done?

src/controllers/vdeployment_controller/proof/liveness/resource_match.rs
```diff
@@ -0,0 +1,1516 @@
...
+    let list_req_msg = |msg| lift_state(and!(at_vd_step_with_vd(vd, controller_id, at_step![AfterListVRS]), req_msg_is_pending_list_req_in_flight(vd, controller_id, msg)));
+    // list_req |= \E |msg| list_req_msg(msg)
```
marshtompsxd: You mean `==` here?

src/controllers/vdeployment_controller/proof/liveness/resource_match.rs
```diff
@@ -0,0 +1,1516 @@
...
+    assert forall |req_msg: Message| inv implies #[trigger] spec.entails(list_req_msg(req_msg).leads_to(list_resp)) by {
+        lemma_from_after_send_list_vrs_req_to_receive_list_vrs_resp(vd, spec, cluster, controller_id, req_msg);
+    };
+    // \A |msg| (list_req_msg(msg) ~> list_resp) |= (\E |msg| list_resp_msg(msg)) ~> list_resp
+    leads_to_exists_intro(spec, |req_msg| list_req_msg(req_msg), list_resp);
+    let list_resp_msg = |msg| lift_state(and!(
+        at_vd_step_with_vd(vd, controller_id, at_step![AfterListVRS]),
+        resp_msg_is_pending_list_resp_in_flight_and_match_req(vd, controller_id, msg)
+    ));
+    // list_resp ~> \E |msg| list_resp_msg(msg)
```
marshtompsxd: Outdated comment?

src/controllers/vdeployment_controller/proof/predicate.rs
```diff
@@ -1,47 +1,503 @@
...
+        &&& req_msg_is_get_then_update_req_with_replicas(vd, controller_id, req_msg, n)(s)
+    }
+}
+
+// a weaker version of coherence between local cache and etcd
+// - only need the key to be in etcd and corresponding objects can pass the filter
+// - so current_state_matches can be reached by sending get-then-update request
+// this predicate holds since AfterListVRS state
+// TODO: simplify this may help to reduce flakiness
+pub open spec fn local_state_match_etcd(vd: VDeploymentView, controller_id: int) -> StatePred<ClusterState> {
```
marshtompsxd: I suggest naming it as `local_state_is_consistent_with_etcd`. `match` means something different in our context.

src/controllers/vdeployment_controller/trusted/liveness_theorem.rs
Catoverflow: Yes

src/controllers/vdeployment_controller/proof/liveness/resource_match.rs
Catoverflow: Yes, I forgot to update the comments

src/controllers/vdeployment_controller/proof/predicate.rs
Catoverflow: Maybe coherent is better?

src/controllers/vdeployment_controller/proof/predicate.rs
```diff
@@ -1,47 +1,503 @@
...
+// a weaker version of coherence between local cache and etcd
+// - only need the key to be in etcd and corresponding objects can pass the filter
+// - so current_state_matches can be reached by sending get-then-update request
+// this predicate holds since AfterListVRS state
+// TODO: simplify this may help to reduce flakiness
+pub open spec fn local_state_is_consistent_with_etcd(vd: VDeploymentView, controller_id: int) -> StatePred<ClusterState> {
+    |s: ClusterState| {
+        let vds = VDeploymentReconcileState::unmarshal(s.ongoing_reconciles(controller_id)[vd.object_ref()].local_state).unwrap();
+        &&& forall |i| 0 <= i < vds.old_vrs_list.len() ==> {
+            // the get-then-update request can succeed
```
marshtompsxd, line 274: To make get-then-update succeed, you need to talk about `etcd_obj`'s namespace and owner_referecnes, not just `vds.old_vrs_list[i]`

src/controllers/vdeployment_controller/proof/predicate.rs
```diff
@@ -1,47 +1,503 @@
...
+            // obj in etcd exists and is owned by vd
+            &&& s.resources().contains_key(vds.old_vrs_list[i].object_ref())
+            &&& ({
+                let etcd_obj = s.resources()[vds.old_vrs_list[i].object_ref()];
+                let etcd_vrs = VReplicaSetView::unmarshal(etcd_obj).unwrap();
+                // can pass list_vrs_obj_filter
+                &&& etcd_obj.kind == VReplicaSetView::kind()
+                &&& etcd_obj.metadata.namespace == vd.metadata.namespace
+                // can pass objects_to_vrs_list
+                &&& VReplicaSetView::unmarshal(etcd_obj).is_ok()
```
marshtompsxd, line 287: You don't have to carry `VReplicaSetView::unmarshal(etcd_obj).is_ok()` in this predicate. There should be an invariant stating that all VRS objects in etcd can be unmarshalled correctly.

marshtompsxd: If you want to pass `objects_to_vrs_list`, it's probably better to assert in the postcondition of the list-resp step that each element in the list response is unmarshallable.

src/controllers/vdeployment_controller/proof/predicate.rs
```diff
@@ -1,47 +1,503 @@
...
+            &&& #[trigger] vds.old_vrs_list[i].well_formed()
+            &&& #[trigger] vds.old_vrs_list[i].metadata.namespace == vd.metadata.namespace
+            &&& #[trigger] vds.old_vrs_list[i].metadata.owner_references_contains(vd.controller_owner_ref())
+            // obj in etcd exists and is owned by vd
+            &&& s.resources().contains_key(vds.old_vrs_list[i].object_ref())
+            &&& ({
+                let etcd_obj = s.resources()[vds.old_vrs_list[i].object_ref()];
+                let etcd_vrs = VReplicaSetView::unmarshal(etcd_obj).unwrap();
+                // can pass list_vrs_obj_filter
+                &&& etcd_obj.kind == VReplicaSetView::kind()
```
marshtompsxd, line 284: Not sure whether you need to state `etcd_obj.kind == VReplicaSetView::kind()` here since we have proved that each object in etcd is well formed

src/controllers/vdeployment_controller/trusted/liveness_theorem.rs
```diff
@@ -20,20 +20,35 @@ pub open spec fn vd_eventually_stable_reconciliation_per_cr(vd: VDeploymentView)
...
-        let filtered_vrs_list = filter_vrs_list(vd, vrs_list.unwrap());
-        let (new_vrs, old_vrs_list) = filter_old_and_new_vrs(vd, filtered_vrs_list);
-        &&& vrs_list.is_Some()
+        // make it consistent with API server's handle_list_req
+        let objs = s.resources().values().filter(list_vrs_obj_filter(vd)).to_seq();
+        let (new_vrs, old_vrs_list) = filter_old_and_new_vrs_on_etcd(vd, s.resources());
+        // this step may return None so we need to check here
+        &&& objects_to_vrs_list(objs).is_Some()
         &&& old_vrs_list.len() == 0
+        // TODO: add requirements on owner_ref
```
marshtompsxd, line 29: What's the requirement on owner_ref? Why not add it now?

marshtompsxd:
@Catoverflow For `local_state_is_consistent_with_etcd`, I think what you need to state is that:
- Each element in `old_vrs_list` is well-formed, has the right namespace, and owner references
- Each element in `old_vrs_list` exists in etcd
- Each element in `old_vrs_list` and the corresponding object in etcd have the same owner references (so get-then-update can apply); you could also say that they have the same `spec` field, but I'm not sure if that's needed
- Each element in `old_vrs_list` and the corresponding object in etcd are indeed *old* vrs by `filter_old_and_new_vrs`
- No other vrs objects in etcd can be seen as *old* vrs by `filter_old_and_new_vrs` (or, all the old vrs are in `old_vrs_list`)

Similarly for new vrs.

Catoverflow:
@marshtompsxd I think it may be a better idea to push changes in `local_state_is_consistent_with_etcd` in the followup PR because the current one is already huge mainly updates top level proof while that predicate is associated with transition lemma.

PR #665, opened 2025-07-09
Finish partial transition lemmas for VDeployment controller

Catoverflow:
- Restructure `req_msg_is_X` predicate
- Update `local_state_is_valid_and_coherent`, we may need to split it up in the future
- Add `vrs_eq_for_vd` as weakened equality when uid and RV are not available
- Add `ObjectMeta.namespace_eq` as a workaround for
  > `error: The verifier does not yet support the following Rust feature: ==/!= for non smt equality types`

### DIFF
--- a/src/controllers/vdeployment_controller/proof/helper_lemmas.rs
+++ b/src/controllers/vdeployment_controller/proof/helper_lemmas.rs
@@ -9,7 +9,7 @@ use crate::temporal_logic::{defs::*, rules::*};
 use crate::vdeployment_controller::{
     model::{install::*, reconciler::*},
     proof::{helper_invariants, predicate::*},
-    trusted::{rely_guarantee::*, spec_types::*, util::*, liveness_theorem::*},
+    trusted::{rely_guarantee::*, spec_types::*, util::*},
 };
 use crate::vreplicaset_controller::trusted::spec_types::*;
 use crate::vstd_ext::{map_lib::*, seq_lib::*, set_lib::*};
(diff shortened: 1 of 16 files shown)


Catoverflow:
@marshtompsxd I want to continue some discussions from #662:

> For `local_state_is_consistent_with_etcd`, I think what you need to state is that:
> 
> ...
>
>  * No other vrs objects in etcd can be seen as _old_ vrs by `filter_old_and_new_vrs` (or, all the old vrs are in `old_vrs_list`)
> 
> 
> Similarly for new vrs.

This doesn't hold during the phase when a `get_then_update_req` has been issued but has not been applied yet. Hence it's kicked out in local state while still present in etcd.

Catoverflow:
And for

> Hence it's kicked out in local state while still present in etcd.

I thought we'd involve some loop in reconciler so I don't keep track of old vrs which has been scaled down in local state. I double checked that and it's possible to keep a list of all old vrs in local state that need to be, or have been scaled down. The solution is similar to VRS controller's by carrying a `diff` in step enum as the index of the next old vrs to be scaled down. This make the reasoning easier as we don't need to look back to previous state to get the vrs being scaled down as it's already kicked out in `s_prime`. But currently that has not caused any problem. If it becomes a obstacle in the future I will change the way to carry old vrs info in local state.

marshtompsxd:
> If it becomes a obstacle in the future I will change the way to carry old vrs info in local state.

Yes, please do so if needed

src/controllers/vdeployment_controller/proof/helper_lemmas.rs
```diff
@@ -189,6 +190,17 @@ pub proof fn only_interferes_with_itself_equivalent_to_lifted_only_interferes_wi
     );
 }
 
+pub proof fn make_replica_set_makes_valid_owned_object(vd: VDeploymentView) -> (vrs: VReplicaSetView)
```
marshtompsxd: Why does this proof need to return a vrs? Why not just have `valid_owned_object(make_replica_set(vd), vd),`

Catoverflow: I think both ways are ok, let me update it

PR #664, opened 2025-07-08
Update Verus version to Release/0.2025.07.05.8bd7c32

Catoverflow:
Update verus to `release/0.2025.07.05.8bd7c32`

Syntax changes:

- `is_Some()` -> `is Some`
- `get_Some_0()` -> `->0`
  - However, `->0` doesn't work in macro in `src/reconciler/spec/io.rs`. I will open an issue in Verus repo later.
- `is_Ok()` -> `is Ok`
- `get_Ok_0()` -> `->Ok_0`
- `is_Err()` -> `is Err`
- `is_None()` -> `is None`

- Use `broadcast group group_seq_properties` instead of `lemma_seq_properties`.
  - However, there's a case in `src/vstd_ext/vec_lib.rs` where `group_seq_properties` didn't work. I will also report it to upstream.

PR #668, opened 2025-07-10
Replace `get_Some_0` with `unwrap` in macro

Catoverflow:
in #664 I replaced most of `get_Some_0` with `->0`, while those in macros defined in `src/reconciler/spec/io.rs` caused some error. As `unwrap` is not deprecated this PR replace them with `unwrap` to reduce the lengthy deprecation warning in every verification.

However, `is_Some` in macro cannot be replaced in this way and will still cause warning. But it's less overwhelming with the `unwrap` replacement.


marshtompsxd:
> However, is_Some in macro cannot be replaced in this way and will still cause warning.

So you cannot do `is_some` in the macro?

PR #669, opened 2025-07-11
Add requirements that `map` and `seq` should be finite

Catoverflow:
Add finite requirement `.metadata.labels` and `.spec.selector.match_labels` for VDeployment and VReplicaSet CRs


marshtompsxd:
@Catoverflow Can you remove the assume?

Catoverflow:
@marshtompsxd I reverted the check in `ObjectMeta`
```rust
    pub open spec fn well_formed_for_namespaced(self) -> bool {
        //..
        &&& self.labels is Some ==> self.labels->0.dom().finite()
        &&& self.annotations is Some ==> self.annotations->0.dom().finite()
    }
```

Because it will break the proof for `etcd_object_is_weakly_well_formed`, which requires add `finite` requirements to other CR's `state_validation`. In current structure I think it's best to just put them in `state_validation` for VD and VRS CRs.

PR #677, opened 2025-07-22
Update the reconciler to update new vrs upon response

Catoverflow:
This is required by coherence predicate:
```rust
VReplicaSetView::unmarshal(s.resources()[new_vrs.object_ref()])->Ok_0 == new_vrs
```
We can have a workaround to exclude `resource_version` and `uid`, but that requires more time to fix than this update on reconciler


Catoverflow:
Closed as is not necessary

PR #683, opened 2025-08-11
Prove lemma_api_request_other_than_pending_req_msg_maintains_objects_owned_by_vd

Catoverflow:
Prove `lemma_api_request_other_than_pending_req_msg_maintains_objects_owned_by_vd`. This lemma picks out the proof at https://github.com/anvil-verifier/anvil/blob/e5d5c53b8c7330183d2014370a849de4a355bcc3/src/controllers/vdeployment_controller/proof/liveness/resource_match.rs#L1517-L1596

And is helpful to prove other API action lemmas

### DIFF
--- a/src/controllers/vdeployment_controller/proof/liveness/api_actions.rs
+++ b/src/controllers/vdeployment_controller/proof/liveness/api_actions.rs
@@ -117,6 +117,9 @@ ensures
     local_state_is_valid_and_coherent(vd, controller_id)(s_prime),
 {}
 
+// This lemma proves for all objects owned by vd (checked by namespace and owner_ref),
+// the API request msg does not change or delete the object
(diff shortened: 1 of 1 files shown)


src/controllers/vdeployment_controller/proof/liveness/api_actions.rs
```diff
@@ -116,4 +117,132 @@ ensures
     local_state_is_valid_and_coherent(vd, controller_id)(s_prime),
 {}
 
+pub proof fn lemma_api_request_other_than_pending_req_msg_maintains_objects_owned_by_vd(
```
marshtompsxd, line 123: @Catoverflow Since the lemma can be so useful, can you write more comments explaining (1) what this lemma proves, and (2) how you prove it (this can be comments embedded inside the proof).

PR #696, opened 2025-10-02
Use `cargo verus` and cargo-native project structure

### LINKED ISSUE #697: Support for out-of-the-box builds with `cargo verus` and crates.io
Hi, 

I discovered Verus & Anvil yesterday, pretty curious about these projects but brand new to this space of course. 

As a newcomer I've found the build process for these projects to be unusual compared to the typical process for Rust applications. However, it seems like there is an ongoing effort to integrate Verus better with `cargo`, see eg this PR at microsoft/verified-storage, where they removed their deps-hack crate and now support the recently added `cargo verus` command: https://github.com/microsoft/verified-storage/pull/34

I made a draft PR to make the `Cargo.toml` point to crates.io, so it does not depend on a local Verus installation (see: https://github.com/anvil-verifier/anvil/pull/696). I was then able to build and verify the project using `cargo verus build`. However, to really integrate with Cargo fully, we would of course need to do more:
* Move the controllers to `src/bin`, [as is convention for Cargo binaries](https://doc.rust-lang.org/cargo/guide/project-layout.html) (currently cargo doesn't recognize any of the controller source files)
  * It seems like the controller source code is a bit out of date w.r.t the rest of the project, but I could have just messed up the code while hacking on it. For example `cluster_state_machine` seems not to exist anymore
* Update CI/CD, Dockerfiles, README & build instructions to use `cargo verus`

I wanted to know if this all sounded reasonable. Happy to submit PR's helping with the effort if so, although I can't yet contribute much to any proof/specification code.

suremarc:
Closes #697

This PR refactors Anvil to be more "`cargo`-native":
* Works out of the box with `cargo verus`, no local dependencies needed
* No more `deps_hack`; all dependencies are in the top-level manifest
* Controller implementations are split into separate binaries from the Anvil library.


Catoverflow:
Hi @suremarc , we are picking up this todo now. The code base has stablized but has changed a lot since this PR. Are you still interested? I will create a new branch on current main and start working on it

Catoverflow:
Close as completed in #800, a nice milestone number

PR #749, opened 2026-02-09
Add Rely Conditions for RMQ Controller

NikhilSDate:


src/controllers/rabbitmq_controller/exec/resource/config_map.rs
```diff
@@ -103,7 +103,8 @@ pub fn make_server_config_map_name(rabbitmq: &RabbitmqCluster) -> (name: String)
     requires rabbitmq@.well_formed(),
     ensures name@ == model_resource::make_server_config_map_name(rabbitmq@),
 {
-    rabbitmq.metadata().name().unwrap().concat("-server-conf")
+    let name = rabbitmq.metadata().name().unwrap();
+    "rabbitmq".to_string().concat("-").concat(name.as_str()).concat("-server-conf")
```
marshtompsxd, line 107: Just curious: Why not just ` "rabbitmq-".to_string().concat(name.as_str()).concat("-server-conf")`?

NikhilSDate: Verus actually fails to automatically prove conformance if I do that. Not sure why.

PR #789, opened 2026-04-15
Use the most recent version of Verus in CI

Catoverflow:
This PR updates CI to use the most recent version of Verus, fixes failed proof caused by the update and removed zookeeper because it's not actively maintained and will not be supported by newer version of Rust soon.

After this update, proofs on VStatefulSet controller are much slower. This performance regression is left to future PR to fix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,59 +36,51 @@ jobs:
         run: |
           echo "home_dir=$HOME" >> $GITHUB_ENV
           echo "home_dir=$HOME"
-      - name: Cache Verus
-        id: cache-verus
+      - name: Cache toolchains and dependencies
+        id: cache-toolchain
         uses: actions/cache@v4
         with:
           path: |
             ${{ env.home_dir }}/verus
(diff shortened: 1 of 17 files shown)


Catoverflow:
Error details:
```shell
verus -L dependency=src/deps_hack/target/debug/deps --crate-type=lib --multiple-errors=100 --extern=deps_hack="src/deps_hack/target/debug/libdeps_hack.rlib" src/esr_composition.rs --allow=unused --verify-only-module vdeployment_controller::proof::liveness::resource_match --verify-function lemma_from_after_send_get_then_update_req_to_receive_ok_resp_of_new_replicas
note: verifying module vdeployment_controller::proof::liveness::resource_match, function esr_composition::vdeployment_controller::proof::liveness::resource_match::lemma_from_after_send_get_then_update_req_to_receive_ok_resp_of_new_replicas (selected functions)

error: variable shadowing not yet supported
 --> vstd/std_specs/option.rs:167:10

error: aborting due to 1 previous error
```

PR #790, opened 2026-04-16
Fix performance regression

Catoverflow:
This PR fixes performance regression caused by updating the version of Verus in #789


Catoverflow:
Proofs of RabbitMQ and VStatefulSet controller are still very slow. I will add another patch later.

PR #791, opened 2026-04-17
Refactor composition

NikhilSDate:


src/controllers/composition/mod.rs
marshtompsxd, line 1: Why comment out these?

NikhilSDate: Since the definition of `ControllerSpec` has changed, it breaks these modules. Since we're going to rewrite the composition proofs using the new framework anyway, I thought it wasn't worth it to try to fix these.

marshtompsxd:
lgtm

PR #794, opened 2026-04-19
Use the most recent Verus release in controller build CI

Catoverflow:
This PR updates CI to use rustup 1.95.0 and controller build workflow


Catoverflow:
https://github.com/verus-lang/verus/pull/2337 is not in release yet, updating CI to build from source

PR #801, opened 2026-06-04
Adapt to finite sets and maps

jaylorch:
Verus will be adding finite sets and maps, in https://github.com/verus-lang/verus/pull/2486. This PR is to prepare for that future.


Catoverflow:
@jaylorch Thanks for the update! I think most of the maps/sets in Anvil are conceptually finite, and the rest can be both. @marshtompsxd could you help confirm?

jaylorch:
Ah, interesting. For simplicity, since you have a huge amount of code :-), I just made everything an `ISet`/`IMap`. But, once the finite sets and maps are available in Verus, you may want to change many of those to `Set` and `Map`.

Or, of course, you could create an alternate PR to this one. You would then only have to change a few of the sets and maps to use `ISet` and `IMap`, and make sure to use constructors for others that ensure that the verifier can verify finiteness. I won't mind if you do that and don't need this PR. :-)

marshtompsxd:
Thank you @jaylorch ! Yes, almost all Map/Set used in Anvil should be finite, so we can choose to use Verus' new Map/Set or IMap/ISet for most cases, depending on which one is easier to port.

@Catoverflow Could you look into the CI failures?

marshtompsxd:
oh I just realized that we will need to upgrade Verus for IMap

jaylorch:
It's expected that CI should fail. Verus hasn't been updated yet. That's why I marked the PR as draft for now.

Catoverflow:
@jaylorch I think we can safely replace all maps in Anvil with finite maps, I can open another PR for this and repair all proofs if you don't mind. After https://github.com/verus-lang/verus/pull/2486 is merged I will immediately merge this PR or that one. Thank you for your effort on finite map support!

jaylorch:
I certainly don't mind; I really appreciate you saving me the effort! Thanks!

Catoverflow:
I opened another PR #802 to adapt to this change in upstream. Some fixes are not trivial because we need to carefully remove finiteness invariants. Thanks again @jaylorch

jaylorch:
On a temporary basis, verita will be pointing at this branch. When you've merged this PR or created your own, please update verita back. (I say "will be" because we accidentally pointed it at the wrong branch to start; we'll fix that shortly.)

https://github.com/verus-lang/verus/blob/main/tools/verita/run_configuration_all.toml

Catoverflow:
Sorry for blocking the update. I will have it ready soon.

jaylorch:
You're not blocking it at all, don't worry.

PR #809, opened 2026-07-01
Use verus tla

Catoverflow:
This PR migrates Anvil to use https://github.com/anvil-verifier/verus-tla


Catoverflow:
I see some flakiness in the proof, will fix it

PR #820, opened 2026-07-30
Proof stabilization needed for the upgrade to Z3 4.16.0

parno:
Please don't merge until [Verus PR 2578](https://github.com/verus-lang/verus/pull/2578) lands.


Catoverflow:
@parno Thank you for the patch! Actually I think the change also works with the old Z3 version and can be merged if CI passes.

And currently only 1/4 Anvil is included in Verita because it takes ~30m to verify the entire project on Github CI machine. Is it possible to parallelize nightly Verita runs? Like having multiple per-project configuration and in Verita CI parallel jobs are summoned accordingly.

Catoverflow:
Oh I need to drop `cargo.lock`, will fix this

Catoverflow:
I opened a PR to fix this problem at https://github.com/parno/anvil/pull/1, @parno could you merge it to unblock this PR? Thank you!

parno:
Done.  Thanks for the patch!

Catoverflow:
You are welcome! Oops it seems there are another 23 proofs broken, because only the proofs for VDeplotment controller are covered. I will send another PR to fix this

Catoverflow:
@parno Sorry but could you also merge https://github.com/parno/anvil/pull/3/? I fixed other 23 rlimit errors for other 3 controllers. A possible solution could be including full verification in Verita config, but as it takes 30m in Github CI a better machine/parallelized verification CI runs may be required.

PR #822, opened 2026-08-19
Verify VSTS get ordinal

Catoverflow:
This PR removes trusted `get_ordinal` and consequently we have to talk about `is_ascii` and:
- add `is_ascii_chars` to `well_formed_for_namespaced`
- add ASCII char check of object name to both model and executable API server when validating a request
- add trusted parser `parse_usize` from string to int, which additionally forbid `000` and strings whose integer values exceed `u32:MAX`
- add axiom `int_to_string_view_ascii`

Some flaky proofs broke and were fixed with adding more rlimit/decomposition into sub-lemmas/`hide(is_ascii_chars)`


src/controllers/vstatefulset_controller/trusted/reconciler.rs
```diff
@@ -2,27 +2,92 @@ use crate::kubernetes_api_objects::exec::prelude::*;
...
+                        seq_equal_preserved_by_add_prefix(prefix@, tail@, int_to_string_view(ord as int));
+                    }
+                }
+            },
+        }
+        ordinal
     }
 
+    // TODO: verify this function
     #[verifier(external_body)]
```
marshtompsxd, line 88: This one is still not verified. Add the todo back?

src/lib.rs
```diff
@@ -22,3 +22,29 @@ pub mod vstatefulset_controller;
 pub mod rabbitmq_controller;
 #[path = "controllers/composition/mod.rs"]
 pub mod composition;
+
```
marshtompsxd: Why so much space?

src/controllers/vstatefulset_controller/trusted/reconciler.rs
Catoverflow: It's added?

src/lib.rs
Catoverflow: lol I forgot what happened...

